PR #40, opened 2019-12-07
Created initial class for `GraphNode`

raajtilaksarma:
As of now, i see `GraphNode` has to have two compulsory variables `key` and `data`. 
Added another member : a list of adjacent vertices. Whenever an edge/vertice is added, we can call any of the two required functions in the `GraphNode` class which are `add_adjacent_key` and `remove_adjacent_key` respectively. Please review this and tell me if there is any other function or member we can add.

### DIFF
--- a/pydatastructs/utils/misc_util.py
+++ b/pydatastructs/utils/misc_util.py
@@ -40,3 +40,30 @@ class TreeNode(object):
         Used for printing.
         """
         return str((self.left, self.key, self.data, self.right))
+
+
+class GraphNode(object):
+    """
+    Represents node in Graphs.
+
+    Parameters
+    ==========
+
+    data
+        Any valid data to be stored in the node.
+    key
+        Required for comparison operations.
+    """
+
+    __slots__ = ['key', 'data']
+
+    def __new__(cls, key, data):
+        obj = object.__new__(cls)
+        obj.data, obj.key = data, key
+        return obj
+
+    def __str__(self):
+        """
+        Used for printing.
+        """
+        return str((self.key, self.data))
--- a/pydatastructs/utils/misc_util.py
+++ b/pydatastructs/utils/misc_util.py
@@ -1,13 +1,13 @@
 from __future__ import print_function, division
 
 __all__ = [
-    'Node'
+    'TreeNode'
 ]
 
 _check_type = lambda a, t: isinstance(a, t)
 NoneType = type(None)
 
-class Node(object):
+class TreeNode(object):
     """
     Represents node in trees.
 
--- a/pydatastructs/trees/__init__.py
+++ b/pydatastructs/trees/__init__.py
@@ -6,7 +6,7 @@ from . import (
 )
 
 from .binary_trees import (
-    Node, BinaryTree, BinarySearchTree, BinaryTreeTraversal, AVLTree
+    TreeNode, BinaryTree, BinarySearchTree, BinaryTreeTraversal, AVLTree
 )
 __all__.extend(binary_trees.__all__)
 
--- a/pydatastructs/trees/binary_trees.py
+++ b/pydatastructs/trees/binary_trees.py
@@ -1,5 +1,5 @@
 from __future__ import print_function, division
-from pydatastructs.utils import Node
+from pydatastructs.utils import TreeNode
 from pydatastructs.miscellaneous_data_structures import Stack
 from pydatastructs.linear_data_structures import (
     OneDimensionalArray, DynamicOneDimensionalArray)
@@ -23,7 +23,7 @@ class BinaryTree(object):
 
     root_data
         Optional, the root node of the binary tree.
-        If not of type Node, it will consider
+        If not of type TreeNode, it will consider
         root as data and a new root node will
         be created.
     key
@@ -53,10 +53,10 @@ class BinaryTree(object):
         if key == None and root_data != None:
             raise ValueError('Key required.')
         key = None if root_data == None else key
-        root = Node(key, root_data)
+        root = TreeNode(key, root_data)
         root.is_root = True
         obj.root_idx = 0
-        obj.tree, obj.size = ArrayForTrees(Node, [root]), 1
+        obj.tree, obj.size = ArrayForTrees(TreeNode, [root]), 1
         obj.comparator = lambda key1, key2: key1 < key2 \
                         if comp == None else comp
         obj.is_order_statistic = is_order_statistic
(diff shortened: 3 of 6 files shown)


czgdp1807:
Thanks for the PR. Please merge `master` into `rts_kwoc`. This is a short cut. I would prefer to have a new branch(checkout from the latest master) for this PR. You can cherry-pick your last commit from this branch and apply on the new branch.

pydatastructs/utils/misc_util.py
```diff
@@ -40,3 +40,30 @@ def __str__(self):
...
+
+    data
+        Any valid data to be stored in the node.
+    key
+        Required for comparison operations.
+    """
+
+    __slots__ = ['key', 'data']
+
+    def __new__(cls, key, data):
```
czgdp1807: I think allowing adjacency list in the constructor of `GraphNode` would be good. Something like,
```python
def __new__(cls, key, data, adj_list):
   ...
```

pydatastructs/utils/misc_util.py
```diff
@@ -40,3 +40,30 @@ def __str__(self):
...
+        obj = object.__new__(cls)
+        obj.data, obj.key = data, key
+        return obj
+
+    def __str__(self):
+        """
+        Used for printing.
+        """
+        return str((self.key, self.data))
```
czgdp1807: Is it a good idea to print adjacency list as well? What do you say?

czgdp1807:
I think add and remove methods should be added to graph. Like edges are usually added to graphs and it would be natural to have `add_edge` and `remove_edge` methods in graphs. Something like,
```python
class Graph:
    ...
    def add_edge(u, v):
        ...
        u.adj_list.append(v)
        ...
```

pydatastructs/utils/misc_util.py
raajtilaksarma: Yes can do that (will be better if we let the user know it is an adjacency list), not just print the list as third argument.

raajtilaksarma:
> Thanks for the PR. Please merge `master` into `rts_kwoc`. This is a short cut. I would prefer to have a new branch(checkout from the latest master) for this PR. You can [cherry-pick](https://git-scm.com/docs/git-cherry-pick) your last commit from this branch and apply on the new branch.

I am a bit new to stuff like merging branches and cherry picking. Will be very grateful if you can guide me here step by step. Also how to get the latest `master`.

pydatastructs/utils/misc_util.py
raajtilaksarma: Agreed on this.

pydatastructs/utils/misc_util.py
czgdp1807: > will be better if we let the user know it is an adjacency list

How can that be done? Using `dict`?

czgdp1807:
Following the steps below should work,
1. git checkout `graph-node` # if you are currently in `master`
2. git cherry-pick f885597

Note that before doing the above pull all the changes locally by executing, `git pull origin master`, where origin if https://github.com/codezonediitj/pydatastructs/

raajtilaksarma:
> Following the steps below should work,
> 
>     1. git checkout `graph-node` # if you are currently in `master`
> 
>     2. git cherry-pick [f885597](https://github.com/codezonediitj/pydatastructs/commit/f88559790eb097031d1201b5aac3843fcf78fefe)
> 
>     3. Make a new PR. Apply the reviews from here.
> 
> 
> Note that before doing the above pull all the changes locally by executing, `git pull origin master`, where origin if https://github.com/codezonediitj/pydatastructs/

Apologize for stretching this. 
```
origin	git@github.com-personal:raajtilaksarma/pydatastructs.git (fetch)
origin	git@github.com-personal:raajtilaksarma/pydatastructs.git (push)
upstream	https://github.com/codezonediitj/pydatastructs (fetch)
upstream	https://github.com/codezonediitj/pydatastructs (push)
```

This is the output of `git remote -v`. I have followed [this](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork) to update my master.

Now i have a branch `rts_kwoc` with latest commits. Should i follow from step 2. in the above comment?

czgdp1807:
I would suggest to make a new PR from a new branch. Cherry pick f885597 and apply on that branch.

pydatastructs/utils/misc_util.py
raajtilaksarma: Yes something like `{'adjacent_nodes' : {key1, key2, key3,. , . , . }}`. The value of the `dict` can be a `set`. I think there can be more than 1 edge between 2 nodes right? So when edges are added we want to add the adjacent nodes in the `set`.

czgdp1807: > I think there can be more than 1 edge between 2 nodes

Can you give an example of such a situation?

I think rather than using `set` we should use `DynamicOneDimensionalArray` as there are many graph algorithms which remove edges, in that case `DynamicOneDimensionalArray` will guarantee amortized constant time.

raajtilaksarma:
> I would suggest to make a new PR from a new branch. Cherry pick [f885597](https://github.com/codezonediitj/pydatastructs/commit/f88559790eb097031d1201b5aac3843fcf78fefe) and apply on that branch.

Output of `f885597 `:
```
On branch graph-node
You are currently cherry-picking commit f885597.

nothing to commit, working tree clean
The previous cherry-pick is now empty, possibly due to conflict resolution.
If you wish to commit it anyway, use:

    git commit --allow-empty
```

Output of `git log --oneline -4`
```
f885597 (HEAD -> graph-node, origin/rts_kwoc, rts_kwoc) graph node class initiated
602a4f5 renamed class Node to TreeNode
dc308d8 (origin/master, origin/HEAD) updated readme (#33)
c0fdee9 Added order statistics to BST (#31)
```

pydatastructs/utils/misc_util.py
raajtilaksarma: https://en.wikipedia.org/wiki/Multigraph
I was talking about this. If we are only working with simple graphs, then we will always have 1 edge.

czgdp1807: Well, in the case of multigraphs we may need a create a separate edge set(a multiset) and adjacency node of Graph node is of very little use in that case. Can you share some use cases of multigraphs, like problems from books/papers/university lectures, so that we can think of how concretely we can represent them?

raajtilaksarma: http://mathworld.wolfram.com/KoenigsbergBridgeProblem.html 
You can take a look at this. I think for now we can work for simple graphs. Usually multi graph problems are related to those with weighted edges mostly. Like finding shortest paths etc.

czgdp1807: Multigraphs may need a different discussion. The current structure of `GraphNode` will be compatible with that. We can keep adjacency list empty if needed in multigraphs. 
Can you raise an issue for crearting MultiGraphs? We can work on it after graphs are added.

raajtilaksarma:
Can we continue the discussion in PR #41

czgdp1807:
Closing this in favor of #40

PR #51, opened 2019-12-18
Added Linked Lists in linear-data-structures

rohansingh9001:
Added linked-lists.py in linear-data-structures
Contains Doubly Linked List

#### References to other Issues or PRs or Relevant literature
Partially fixes issue #49 which requests to add linked lists.
Only Doubly Linked List added so far Single Linked List remains.

#### Brief description of what is fixed or changed
API for the additions are as follows -

API for Double Linked List:

1)A Doubly linked list should be initialised with `DoubleLinkedList()`

2)Elements can be pushed at the start of the list using `.appendleft(data)`

3)Elements can be appended at the end using `.appendright(data)`

4) can be popped from the list at a certain index by `.pop(index)`

5)`__getitem__` and `__setitem__` support using python syntax,
eg -
`list[1]`
`list[5] =4`

6)Print the Linked List in List form using `print()` function
eg -
if elements in linked list named `sample_list` are 1-2-3-4
`>>>print(sample_list)`
`[1,2,3,4]`

7)Return length of the Linked List using `len()` function.

8)`.popright()` and `.popleft()` pops items from the right and the left respectively.

9)`.insertAt(index, data)` inserts a new Node at the index.

10)`.insertBefore(Node, data)` and `.insertAfter(Node, data)` inserts a new Node before and after the input Node respectively.

PR #54, opened 2019-12-20
Test Using 'Raises' are now asserted!

### LINKED ISSUE #52: Tests using `raises` are not being asserted

I was observing the test files and found that tests which use `raises` are not being asserted which was due to absence of `return` statement in, 
https://github.com/codezonediitj/pydatastructs/blob/8da8f36b1243a9acd0105c422baccc05831da936/pydatastructs/utils/raises_util.py#L3-L16

However with merging of https://github.com/codezonediitj/pydatastructs/pull/51 `raises` will return `True` on successful execution, hence all the tests of the following format,
```python
raises(...)
``` 
are to be replace with the format below,
```python
assert raises(...)
```

TarunTomar122:
Fixes #52 by asserting the raises.

changed all the raises(.....) to 
assert raises(.....)

### DIFF
--- a/pydatastructs/linear_data_structures/tests/test_arrays.py
+++ b/pydatastructs/linear_data_structures/tests/test_arrays.py
@@ -11,14 +11,14 @@ def test_OneDimensionalArray():
     assert ODA(int, [1.0, 2, 3, 4, 5], 5)
     assert ODA(int, 5)
     assert ODA(int, [1.0, 2, 3])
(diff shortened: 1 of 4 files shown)


czgdp1807:
Thanks for the PR. The tests will fail for now. If @rohansingh9001 doesn't complete https://github.com/codezonediitj/pydatastructs/pull/51 by tomorrow then I will make a follow up PR.

PR #63, opened 2019-12-26
Updated object to class name #57

### LINKED ISSUE #57: `obj = object.__new__(cls)`

Currently any class object is created by,
```python
obj = object.__new__(cls)
```
For better readability, each such line should be replaced by the abstract superclass of that class.
Something like,
```python
obj = <name_of_abstract_super_class>.__new__(cls)
```
For example, in `arrays.py`,  `OneDimensionalArray` is created by, 
https://github.com/codezonediitj/pydatastructs/blob/9c32d4068ace96da9b953ea678b1c2fd852d141c/pydatastructs/linear_data_structures/arrays.py#L73
However, the above should be replaced by,
```python
obj = Array.__new__(cls)
```

TarunTomar122:
fixes #57

Replaced lines having object.__new__ to base class name __new__..

### DIFF
--- a/pydatastructs/miscellaneous_data_structures/stack.py
+++ b/pydatastructs/miscellaneous_data_structures/stack.py
@@ -96,7 +96,7 @@ class ArrayStack(Stack):
         if items._size > maxsize:
             raise ValueError("Overflow, size of items %s is greater "
                             "than maxsize, %s"%(items._size, maxsize))
-        obj = Stack.__new__(cls)
+        obj = object.__new__(cls)
         obj.maxsize, obj.top, obj.items, obj.dtype = \
             maxsize, top, items, items._dtype
         return obj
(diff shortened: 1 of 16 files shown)


pydatastructs/miscellaneous_data_structures/stack.py
```diff
@@ -96,7 +96,7 @@ def __new__(cls, maxsize=None, top=0, items=None, dtype=int):
         if items._size > maxsize:
             raise ValueError("Overflow, size of items %s is greater "
                             "than maxsize, %s"%(items._size, maxsize))
-        obj = object.__new__(cls)
+        obj = Stack.__new__(cls)
```
czgdp1807: We have to keep it as `object`. This is an exception to the generic fix. Rest is fine.

TarunTomar122: Done!!

czgdp1807:
Make the changes at the following place as well,

https://github.com/codezonediitj/pydatastructs/blob/27d3ed84602f157e889ad7d3e25a10af10750700/pydatastructs/trees/heaps.py#L70

TarunTomar122:
But Binary heap is not inheriting any other class than object class itself!

czgdp1807:
Well, then we need to make it inherit, `Heap` and then make the change. In future, there may be a need to shift some common methods to the topmost class.

czgdp1807:
Thanks. Merging.

PR #83, opened 2020-01-15
Fixed Incorrectness in Documentation

### LINKED ISSUE #79: Incorrectness in Documentation

There are some places in the documentation where it is said that `Stack` to implement iterative version of recursive algorithms but in the code `Queue` is used. A fix is needed in the documentation and it should be mentioned that `Queue` is used instead of `Stack`.

The task is to search the entire documentation(which is not too large at the time of the creation of this issue) and correct it. No change in code is required.

iamrajiv:

Fixes #79

It fixes the incorrectness in documentation i.e. earlier there were some places in the documentation where it is said that `Stack` to implement an iterative version of recursive algorithms but in the code `Queue` is used. So this can be fixed by mentioning that `Queue` is used instead of `Stack`

#### Other comments

### DIFF
--- a/pydatastructs/trees/space_partitioning_trees.py
+++ b/pydatastructs/trees/space_partitioning_trees.py
@@ -122,7 +122,7 @@ class OneDimensionalSegmentTree(object):
     def build(self):
         """
         Builds the segment tree from the segments,
-        using iterative algorithm based on stacks.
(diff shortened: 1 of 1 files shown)


czgdp1807:
LGTM. Merging.

Issue #107, opened 2020-03-01
Adding `sort` method to heaps


ajaynair710:

czgdp1807:
It looks like this issue is quite problem specific. Which algorithm and data structure will be used to do this task?

ajaynair710:
heap sort

czgdp1807:
IMO, a `sort` function can be added in `BinaryHeap` for such tasks.

Shweta0002:
Can I work on this?? Interested to participate in RGSoC.

czgdp1807:
Hi, @Shweta0002 I would suggest to add a `sort` function in class `BinaryHeap` or in general for any heap in PyDataStructs. This function might return the nodes of the heap sorted. Please see the code for heaps at https://github.com/codezonediitj/pydatastructs/blob/master/pydatastructs/trees/heaps.py and find a suitable API for the method, `sort`.

Shweta0002:
Oh ok ! But didn't get that API thing ?

czgdp1807:
Basically API in the context of this issue means,  the prototype of the method `sort`. What arguments it should take, other than `self`(`BinaryHeap` object)?

Shweta0002:
ok , so for now i have to add a heap sort function in heap file right?So I would like to do it , thanks.

czgdp1807:
`sort` should be a member function of `Heap` class which can have different implementations in the child classes. You can proceed with a PR for this and rest of the things can be discussed there.

nehasangeetajha:
can we have intro sort as one of the sorting techniques?
If yes can I try it?

czgdp1807:
> intro sort

Can you please provide some reference to this sorting technique?

nehasangeetajha:
@czgdp1807 Here's is it : 
https://en.wikipedia.org/wiki/Introsort

czgdp1807:
We can add all the sorting algorithms in a file named, `algorithms` under `linear_data_structures`. The thing to consider is whether to add sequential version of such sorting algorithms or parallel ones. Optimized sequential sorting algorithms are pre-implemented in many languages/libraries. 
How can we make the introsort parallel? Any ideas?

nehasangeetajha:
Can't say about introsort but would love to start with basic sorting techniques (making them parallel) if we can have this folder

czgdp1807:
[...]

PR #116, opened 2020-03-02
Sorting words in lexicographical order #107

ajaynair710:


czgdp1807:
Closing.  AVL trees are already there in the `master`. The sorting algorithm used in the other file is just a call to inbuilt method of python.  
Thanks for the contributions.

PR #118, opened 2020-03-02
DHeap added

### LINKED ISSUE #102: Add m-ary heap

Currently, the package does not have m-ary heap (a generalization of the binary heap).
https://www.geeksforgeeks.org/k-ary-heap/

prshnt19:

Closes #102

### DIFF
--- a/pydatastructs/linear_data_structures/arrays.py
+++ b/pydatastructs/linear_data_structures/arrays.py
@@ -255,7 +255,6 @@ class DynamicOneDimensionalArray(DynamicArray, OneDimensionalArray):
     def size(self):
         return self._size
 
-
 class ArrayForTrees(DynamicOneDimensionalArray):
     """
     Utility dynamic array for storing nodes of a tree.
--- a/pydatastructs/linear_data_structures/arrays.py
+++ b/pydatastructs/linear_data_structures/arrays.py
@@ -255,6 +255,7 @@ class DynamicOneDimensionalArray(DynamicArray, OneDimensionalArray):
     def size(self):
         return self._size
 
+
 class ArrayForTrees(DynamicOneDimensionalArray):
     """
     Utility dynamic array for storing nodes of a tree.
--- a/pydatastructs/trees/heaps.py
+++ b/pydatastructs/trees/heaps.py
@@ -82,15 +82,15 @@ class DHeap(Heap):
         else:
             raise ValueError("%s is invalid heap property"%(heap_property))
         if elements is None:
-            elements = []
+            elements = DynamicOneDimensionalArray(TreeNode, 0)
         obj.heap = elements
-        obj._last_pos_filled = len(elements) - 1
+        obj._last_pos_filled = obj.heap._last_pos_filled
         obj._build()
         return obj
 
     def _build(self):
         for i in range(self._last_pos_filled + 1):
-            self.heap[i].left, self.heap[i].right = \
+            self.heap[i]._leftmost, self.heap[i]._rightmost = \
                 self.d*i + 1, self.d*i + self.d
         for i in range((self._last_pos_filled + 1)//self.d, -1, -1):
             self._heapify(i)
@@ -143,7 +143,7 @@ class DHeap(Heap):
         self.heap.append(new_node)
         self._last_pos_filled += 1
         i = self._last_pos_filled
-        self.heap[i].left, self.heap[i].right = self.d*i + 1, self.d*i + self.d
+        self.heap[i]._leftmost, self.heap[i]._rightmost = self.d*i + 1, self.d*i + self.d
 
         while True:
             parent = (i - 1)//self.d
@@ -171,22 +171,20 @@ class DHeap(Heap):
         else:
             element_to_be_extracted = TreeNode(self.heap[0].key, self.heap[0].data)
             self._swap(0, self._last_pos_filled)
-            self.heap[self._last_pos_filled] = TreeNode(float('inf') if self.heap_property == 'min'
-                                                                else float('-inf'), None)
-            self._heapify(0)
-            self.heap.pop()
+            self.heap.delete(self._last_pos_filled)
             self._last_pos_filled -= 1
+            self._heapify(0)
             return element_to_be_extracted
 
     def __str__(self):
         to_be_printed = ['' for i in range(self._last_pos_filled + 1)]
         for i in range(self._last_pos_filled + 1):
             node = self.heap[i]
-            if node.left <= self._last_pos_filled:
-                if node.right <= self._last_pos_filled:
-                    children = [x for x in range(node.left, node.right + 1)]
+            if node._leftmost <= self._last_pos_filled:
+                if node._rightmost <= self._last_pos_filled:
+                    children = [x for x in range(node._leftmost, node._rightmost + 1)]
                 else:
-                    children = [x for x in range(node.left, self._last_pos_filled + 1)]
+                    children = [x for x in range(node._leftmost, self._last_pos_filled + 1)]
             else:
                 children = []
             to_be_printed[i] = (node.key, node.data, children)
@@ -303,6 +301,7 @@ class TernaryHeap(DHeap):
     ==========
 
     .. [1] https://en.wikipedia.org/wiki/D-ary_heap
+    .. [2] https://ece.uwaterloo.ca/~dwharder/aads/Algorithms/d-ary_heaps/Ternary_heaps/
     """
     def __new__(cls, elements=None, heap_property="min"):
         obj = DHeap.__new__(cls, elements, heap_property, 3)
--- a/pydatastructs/trees/heaps.py
+++ b/pydatastructs/trees/heaps.py
@@ -5,6 +5,7 @@ from pydatastructs.miscellaneous_data_structures.binomial_trees import BinomialT
 
 __all__ = [
     'BinaryHeap',
+    'TernaryHeap',
     'DHeap',
     'BinomialHeap'
 ]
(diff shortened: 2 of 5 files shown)


prshnt19:
The `_build(self)` in `BinaryHeap` doesn't seem to be doing anything. Its called once when the heap is to be built. At that time heap is empty so nothing actually happens.

czgdp1807:
Well by default, elements can be `None` so `_build` does nothing. However, if you pass some list in elements then `_build` will work.

czgdp1807:
Is it possible to remove the class `BinaryHeap` and add a constructor function with a same protoype as `BinaryHeap.__new__`?
```python
def BinaryHeap(arguments accepted by BinaryHeap.__new__)
```
We can also add a special constructor for `TernaryHeap`.

prshnt19:
I think a better alternative will be to inherit `DHeap`. `BinaryHeap` and `TernaryHeap` will be a child class with d=2 and d=3 respectively.
```
class BinaryHeap(DHeap):
    def __new__(cls, elements=None, heap_property="min"):
         obj = DHeap.__new__(cls, elements=None, heap_property="min", d=2)
         return obj
```


```
class TernaryHeap(DHeap):
    def __new__(cls, elements=None, heap_property="min"):
         obj = DHeap.__new__(cls, elements=None, heap_property="min", d=3)
         return obj
```

prshnt19:
ping @czgdp1807

czgdp1807:
Can we use `DynamicOneDimensionalArray` [here](https://github.com/codezonediitj/pydatastructs/pull/118/files#diff-490d802c52ac67dcfaaf581599aaf045R86)?

pydatastructs/trees/heaps.py
```diff
@@ -180,12 +182,131 @@ def __str__(self):
...
+    >>> max_heap.extract().key
+    7
+    >>> max_heap.insert(6, 6)
+    >>> max_heap.extract().key
+    6
+
+    References
+    ==========
+
+    .. [1] https://en.wikipedia.org/wiki/D-ary_heap
```
czgdp1807: Any references for `TernaryHeap`?

pydatastructs/trees/heaps.py
```diff
@@ -65,13 +67,14 @@ class BinaryHeap(Heap):
     References
     ==========
 
-    .. [1] https://en.m.wikipedia.org/wiki/Binary_heap
+    .. [1] https://en.wikipedia.org/wiki/D-ary_heap
     """
-    __slots__ = ['_comp', 'heap', 'heap_property', '_last_pos_filled']
+    __slots__ = ['_comp', 'heap', 'm', 'heap_property', '_last_pos_filled']
```
czgdp1807:
```suggestion
    __slots__ = ['_comp', 'heap', 'd', 'heap_property', '_last_pos_filled']

```
Let's keep `d` to denote the generalisation everywhere.

pydatastructs/trees/heaps.py
```diff
@@ -180,12 +182,131 @@ def __str__(self):
...
+    >>> max_heap = TernaryHeap(heap_property='max')
+    >>> max_heap.insert(1, 1)
+    >>> max_heap.insert(5, 5)
+    >>> max_heap.insert(7, 7)
+    >>> max_heap.extract().key
+    7
+    >>> max_heap.insert(6, 6)
+    >>> max_heap.extract().key
+    6
+
```
czgdp1807: These examples do not show the working of `TernaryHeap`. The structure will be similar to `BinaryHeap`. May be one more insert would show the difference.

pydatastructs/trees/tests/test_heaps.py
```diff
@@ -47,6 +47,88 @@ def test_BinaryHeap():
...
+    max_heap.insert(2, 2)
+    max_heap.insert(7, 7)
+    assert str(max_heap) == \
+           ('[(100, 100, [1, 2, 3]), (25, 25, [4, 5, 6]), '
+            '(36, 36, [7, 8]), (17, 17, []), '
+            '(3, 3, []), (19, 19, []), (1, 1, []), '
+            '(2, 2, []), (7, 7, [])]')
+
+    expected_extracted_element = max_heap.heap[0].key
+    assert max_heap.extract().key == expected_extracted_element
```
czgdp1807:
```suggestion
    assert max_heap.extract().key == 100

```

pydatastructs/trees/tests/test_heaps.py
```diff
@@ -47,6 +47,88 @@ def test_BinaryHeap():
...
+    for _ in range(8):
+        sorted_elements.append(max_heap.extract().key)
+    assert expected_sorted_elements == sorted_elements
+
+    elements = [
+        TreeNode(7, 7), TreeNode(25, 25), TreeNode(100, 100),
+        TreeNode(1, 1), TreeNode(2, 2), TreeNode(3, 3),
+        TreeNode(17, 17), TreeNode(19, 19), TreeNode(36, 36)
+    ]
+    min_heap = BinaryHeap(elements=elements, heap_property="min")
```
czgdp1807:
```suggestion
    min_heap = TernaryHeap(elements=elements, heap_property="min")

```
May be you wanted the above.

pydatastructs/trees/tests/test_heaps.py
```diff
@@ -47,6 +47,88 @@ def test_BinaryHeap():
...
+    max_heap = DHeap(heap_property="max", d=4)
+    assert max_heap.extract() is None
+    max_heap.insert(100, 100)
+    max_heap.insert(19, 19)
+    max_heap.insert(36, 36)
+    max_heap.insert(17, 17)
+    max_heap.insert(3, 3)
+    max_heap.insert(25, 25)
+    max_heap.insert(1, 1)
+    max_heap = DHeap(max_heap.heap, heap_property="max", d=3)
```
czgdp1807: We should test it for `d=4` as `d=3` and `d=2` have already been tested.

pydatastructs/trees/heaps.py
prshnt19: OK, it was by mistake.

pydatastructs/trees/heaps.py
prshnt19: I can add one more insert. But one can't verify it's `TernaryHeap` heap by insertion unless we add feature like height. But adding feature like height is unnecessary.

pydatastructs/trees/heaps.py
prshnt19: There was no individual reference to `TernaryHeap`. But it was there on the Wiki page of [DHeap](https://en.wikipedia.org/wiki/D-ary_heap)

pydatastructs/trees/tests/test_heaps.py
prshnt19: This line is actually for testing heap_property. 

The actual testing is on d=4 only. And d=3 is used for conversion from one type of heap to other.

prshnt19:
> Can we use `DynamicOneDimensionalArray` [here](https://github.com/codezonediitj/pydatastructs/pull/118/files#diff-490d802c52ac67dcfaaf581599aaf045R86) instead of `list`?

The `DynamicOneDimensionalArray` itself uses `list`. I think using `DynamicOneDimensionalArray` would be a good idea only if we use `cyptes` for `DynamicOneDimensionalArray`. This will give a performance boost. I can create a new issue for that.

pydatastructs/trees/heaps.py
prshnt19: I have updated the references.

czgdp1807:
Though `DynamicOneDimensionalArray` uses list but the deletion time for the same is amortized O(1) with a nice trade off with space complexity. What happens if we use it?

pydatastructs/trees/heaps.py
```diff
@@ -141,10 +143,10 @@ def insert(self, key, data):
         self.heap.append(new_node)
         self._last_pos_filled += 1
         i = self._last_pos_filled
-        self.heap[i].left, self.heap[i].right = 2*i + 1, 2*i + 2
+        self.heap[i].left, self.heap[i].right = self.d*i + 1, self.d*i + self.d
```
czgdp1807: May be, `left` -> `leftmost` and `right` -> `rightmost` will sound better.

prshnt19: But it may sounds odd for `BinaryHeap`.

czgdp1807: Two properties can be defined for `BinaryHeap`,
```python
@property
def left(self):
    return self.leftmost

@property
def right(self):
    return self.rightmost
```

prshnt19: ok

prshnt19:
Used `DynamicOneDimensionalArray` instead of `List` in `DHeap`.  `DynamicOneDimensionalArray` could be made faster using `ctypes` array.

`ctypes` implementations are 10 to 100 times faster than python and takes less memory than python.

czgdp1807:
> `DynamicOneDimensionalArray` could be made faster using `ctypes` array. `ctypes` implementations are 10 to 100 times faster than python and takes less memory than python.

Can you please make an experimental PR for the same?

pydatastructs/linear_data_structures/arrays.py
```diff
@@ -255,6 +255,7 @@ def delete(self, idx):
     def size(self):
         return self._size
 
+
```
czgdp1807:
```suggestion
```

pydatastructs/trees/tests/test_heaps.py
```diff
@@ -20,15 +21,16 @@ def test_BinaryHeap():
...
+        "(3, 3, []), (25, 25, []), (1, 1, []), "
+        "(2, 2, []), (7, 7, [])]")
 
-    expected_extracted_element = max_heap.heap[0].key
-    assert max_heap.extract().key == expected_extracted_element
+    assert max_heap.extract().key == 100
 
     expected_sorted_elements = [36, 25, 19, 17, 7, 3, 2, 1]
+    l=max_heap.heap[0].left
+    l=max_heap.heap[0].right
```
czgdp1807:
```suggestion
    l = max_heap.heap[0].left
    l = max_heap.heap[0].right

```

Issue #113, opened 2020-03-01
Linked list implementation of stack and queue data structure


asaw4:
kindly assign me this problem .

czgdp1807:
Array implementation of stacks and queues have already been added. See, https://github.com/codezonediitj/pydatastructs/tree/master/pydatastructs/miscellaneous_data_structures
Feel free to add the linked list implementation of these two once, #99 is merged which will be done by tomorrow.

abhishekgupta368:
can i contribute

czgdp1807:
@abhishekgupta368 Sure, you can.  We allow only one PR to remain open for any issue. The best PR will be kept and rest others will be closed in its favour.  See our PR policy at https://github.com/codezonediitj/pydatastructs/wiki/Pull-Request-Policy

czgdp1807:
Note that #99 will be merged by tomorrow after which you can start your work on linked list implementation of stacks and queues.

HarsheetKakar:
Can I work on LinkedListStack implementation?

nehasangeetajha:
Is this issue still available? If yes can I start working on this?

HarsheetKakar:
waiting for #152 to be merged, queue is already implemented, stack on its way

PR #127, opened 2020-03-06
LinkedListQueue Added

### LINKED ISSUE #49: Add Linked Lists
Currently the package does not have Queues or Linked Lists. 
While working with issue #22-Add Queue  
It was discovered that to add a Queue which is efficient enough to perform popleft() action in O(1) time we need linked lists to implement Queues just like the collections.deque implementaion of a Queue.

Since the original goal of issue #22 was to reduce dependencies on other libraries, LInked List must be added to pydatastructs to ensure an efficient implementation of Queues.

#### Example of the problem

>>> sample_queue = Queue()
>>> for i in range(N) # N be a large number
...        sample_queue.append('#')
...
>>> sample_queue.popleft() # Takes O(N) time
#

See the original issue -
https://github.com/codezonediitj/pydatastructs/issues/22

Read about the efficiency issue here -
https://www.geeksforgeeks.org/deque-in-python/
https://stackoverflow.com/questions/55578874/dequeue-complexity-on-using-python-list-as-a-queue

prshnt19:

FIxes #49 #113

### DIFF
--- a/pydatastructs/miscellaneous_data_structures/queue.py
+++ b/pydatastructs/miscellaneous_data_structures/queue.py
@@ -1,5 +1,5 @@
-from pydatastructs.linear_data_structures import DynamicOneDimensionalArray
-from pydatastructs.utils.misc_util import NoneType
+from pydatastructs.linear_data_structures import DynamicOneDimensionalArray, SinglyLinkedList
+from pydatastructs.utils.misc_util import NoneType, LinkedListNode
 from copy import deepcopy as dc
 
 __all__ = [
@@ -51,6 +51,12 @@ class Queue(object):
             return ArrayQueue(
                 kwargs.get('items', None),
                 kwargs.get('dtype', int))
+        
+        elif implementation == 'linkedlist':
+            return LinkedListQueue(
+                # kwargs.get('items', None),
+                # kwargs.get('dtype', LinkedListNode)
+            )
         raise NotImplementedError(
                 "%s hasn't been implemented yet."%(implementation))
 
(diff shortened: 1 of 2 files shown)


pydatastructs/miscellaneous_data_structures/queue.py
```diff
@@ -51,6 +51,11 @@ def __new__(cls, implementation='array', **kwargs):
             return ArrayQueue(
                 kwargs.get('items', None),
                 kwargs.get('dtype', int))
+        elif implementation == 'linkedlist':
+            return LinkedListQueue(
+                # kwargs.get('items', None),
+                # kwargs.get('dtype', LinkedListNode)
```
czgdp1807: Why is this commented out? Can we keep the API similar to the `ArrayQueue`?

pydatastructs/miscellaneous_data_structures/queue.py
```diff
@@ -64,7 +69,9 @@ def popleft(self, *args, **kwargs):
 
     @property
     def is_empty(self):
-        return None
+        raise NotImplementedError(
+            "This is an abstract method.")
+
```
czgdp1807: `return None` was used to show the abstractness. I think the change is fine to let the user know they are trying to access something that isn't defined.

prshnt19: Also, it improves Codecov.

pydatastructs/miscellaneous_data_structures/queue.py
prshnt19: Made `LinkedListQueue` analogous to `ArrayQueue`. Now they have similar API.

czgdp1807:
@manaswinidas Please see if the tests are conceptually correct.

Issue #166, opened 2020-03-17
LinkedListNode documentation not complete.


HarsheetKakar:
#### Description of the problem
Missing info on links and addrs parameters.

#### Example of the problem
```
class LinkedListNode(Node):
    """
    Represents node in linked lists.

    Parameters
    ==========

    data
        Any valid data to be stored in the node.
    """
    def __new__(cls, data=None, links=['next'], addrs=[None]):
```

czgdp1807:
Feel free to add the necessary information.

HarsheetKakar:
Honestly speaking I have no idea what links or addrs is doing.

czgdp1807:
`links` is a list containing the names of attributes which should be used as links to other nodes. For example if you want to create a node for doubly linked lists then you can pass, `links = ['next', 'prev']`. `addrs` is contains the address of nodes to be assigned to each of the attributes in `links`. 
This is done for using only a single node class for various kind linked lists with differences in the number of links only.

HarsheetKakar:
why do we have 2 lists for that then? shouldn't it be in a dict maybe?

urbashi123:
I want to work on documetation.Can i work on it?

PR #171, opened 2020-03-18
Skip list added

### LINKED ISSUE #149: Add a Skip List Data Structure
[1] https://en.wikipedia.org/wiki/Skip_list

HarsheetKakar:

Fixes #149 

Highly influenced by https://github.com/TheAlgorithms/Python/blob/a9f73e318cddf43769083614a3e1f9dab1ec50fc/data_structures/linked_list/skip_list.py

SkipList added in linked_list.py

#### Other comments
the __str__() method in SkipList class is a bit different, please check if it is ok.

### DIFF
--- a/pydatastructs/linear_data_structures/linked_lists.py
+++ b/pydatastructs/linear_data_structures/linked_lists.py
@@ -756,7 +756,7 @@ class SkipList(LinkedList):
     ========
 
     >>> from pydatastructs import SkipList
-    >>> skip_list = SkipList()
+    >>> skip_list = SkipList(0.5, 16)
     >>> skip_list.insert(2, "2")
     >>> skip_list.insert(4, "4")
     >>> skip_list.insert(6, "4")
--- a/pydatastructs/linear_data_structures/linked_lists.py
+++ b/pydatastructs/linear_data_structures/linked_lists.py
@@ -743,6 +743,15 @@ class SkipList(LinkedList):
     """
     Represents Skip List
 
+    Parameters
+    ==========
+
+    p: float
+        Probability that node at level i appears in level i+1.
+
+    levels: int
+        Number of levels in the skip list.
+
     Examples
     ========
 
@@ -761,15 +770,6 @@ class SkipList(LinkedList):
 
     .. [1] https://epaperpress.com/sortsearch/download/skiplist.pdf
     .. [2] https://github.com/TheAlgorithms/Python/blob/a9f73e318cddf43769083614a3e1f9dab1ec50fc/data_structures/linked_list/skip_list.py
-
-    Parameters
-    ==========
-
-    p
-        Fraction of nodes of level i which are present at level i+1 also.
-
-    max_level
-        Maximum level that can be attained by any node.
 
     """
 
--- a/pydatastructs/linear_data_structures/linked_lists.py
+++ b/pydatastructs/linear_data_structures/linked_lists.py
@@ -826,7 +826,7 @@ class SkipList(LinkedList):
             yield node.forward[0].key
             node = node.forward[0]
 
-    def random_level(self):
+    def _random_level(self):
         level = 1
         while random() < self.p and level < self.max_level:
             level += 1
@@ -895,7 +895,7 @@ class SkipList(LinkedList):
         if node is not None:
             node.value = value
         else:
-            level = self.random_level()
+            level = self._random_level()
 
             if level > self.level:
                 for i in range(self.level - 1, level):
--- a/pydatastructs/linear_data_structures/linked_lists.py
+++ b/pydatastructs/linear_data_structures/linked_lists.py
@@ -45,6 +45,66 @@ class LinkedList(object):
                 break
         return str(elements)
 
+    def insert_after(self, prev_node, key, data=None):
+        """
+        Inserts a new node after the prev_node.
+
+        Parameters
+        ==========
+
+        prev_node: LinkedListNode
+            The node after which the
+            new node is to be inserted.
+
+        data
+            Any valid data to be stored in the node.
+        """
+        raise NotImplementedError('This is an abstract method')
+
+    def insert_at(self, index, key, data=None):
+        """
+        Inserts a new node at the input index.
+
+        Parameters
+        ==========
+
+        index: int
+            An integer satisfying python indexing properties.
+
+        data
+            Any valid data to be stored in the node.
+        """
+        raise NotImplementedError('This is an abstract method')
+
+    def extract(self, index):
+        """
+        Extracts the node at the index of the list.
+
+        Parameters
+        ==========
+
+        index: int
+            An integer satisfying python indexing properties.
+
+        Returns
+        =======
+
+        current_node: LinkedListNode
+            The node at index i.
+        """
+        raise NotImplementedError('This is an abstract method')
+
+    def __getitem__(self, index):
+        """
+        Returns
+        =======
+
+        current_node: LinkedListNode
+            The node at given index.
+        """
+        raise NotImplementedError('This is an abstract method')
+
+
 class DoublyLinkedList(LinkedList):
     """
     Represents Doubly Linked List
--- a/pydatastructs/linear_data_structures/linked_lists.py
+++ b/pydatastructs/linear_data_structures/linked_lists.py
@@ -785,25 +785,6 @@ class SkipList(LinkedList):
         return obj
 
     def __str__(self):
-        """
-        :return: Visual representation of SkipList
-        >>> skip_list = SkipList(0.5, 16)
-        >>> print(skip_list)
-        SkipList(level=0)
-        >>> skip_list.insert("Key1", "Value")
-        >>> print(skip_list) # doctest: +ELLIPSIS
-        SkipList(level=...
-        [root]--...
-        [Key1]--Key1...
-        None    *...
-        >>> skip_list.insert("Key2", "OtherValue")
-        >>> print(skip_list) # doctest: +ELLIPSIS
-        SkipList(level=...
-        [root]--...
-        [Key1]--Key1...
-        [Key2]--Key2...
-        None    *...
-        """
 
         items = list(self)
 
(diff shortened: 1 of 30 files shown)


czgdp1807:
Can you please make a separate PR for fixing #166 ?

pydatastructs/linear_data_structures/linked_lists.py
```diff
@@ -722,3 +724,213 @@ def extract(self, index):
         elif index == 0:
             self.tail.next = self.head
         return node
+
+class SkipList(LinkedList):
+    """
+    Represents Doubly Circular Linked List
```
czgdp1807:
```suggestion
    Represents Skip List
```

czgdp1807:
> https://github.com/TheAlgorithms/Python/blob/a9f73e318cddf43769083614a3e1f9dab1ec50fc/data_structures/linked_list/skip_list.py

We don't know how their API is compatible with ours. So may be we need some discussion on it.

HarsheetKakar:
> > https://github.com/TheAlgorithms/Python/blob/a9f73e318cddf43769083614a3e1f9dab1ec50fc/data_structures/linked_list/skip_list.py
> 
> We don't know how their API is compatible with ours. So may be we need some discussion on it.

I have tested it and made the changes I deemed fit, still take a look and tell me if there is something you would like changed and I am available 24/7 :) .

HarsheetKakar:
> Can you please make a separate PR for fixing #166 ?

Sure

HarsheetKakar:
Do you want me to change something in it? @czgdp1807

pydatastructs/linear_data_structures/linked_lists.py
```diff
@@ -722,3 +724,213 @@ def extract(self, index):
...
+    .. [2] https://github.com/TheAlgorithms/Python/blob/a9f73e318cddf43769083614a3e1f9dab1ec50fc/data_structures/linked_list/skip_list.py
+
+    Parameters
+    ==========
+
+    p
+        Fraction of nodes of level i which are present at level i+1 also.
+
+    max_level
+        Maximum level that can be attained by any node.
```
czgdp1807:
```suggestion
    Parameters
    ==========
    
    p: float
        Fraction of nodes of level i which are present at level i+1 also.

    max_level: int
        Maximum level that can be attained by any node.

    Examples
    ========

    >>> from pydatastructs import SkipList
    >>> skip_list = SkipList()
    >>> skip_list.insert(2, "2")
    >>> skip_list.insert(4, "4")
    >>> skip_list.insert(6, "4")
    >>> skip_list.insert(4, "5")
    >>> skip_list.insert(8, "4")
    >>> skip_list.insert(9, "4")
    >>> skip_list.delete(4)

    References
    ==========

    .. [1] https://epaperpress.com/sortsearch/download/skiplist.pdf
    .. [2] https://github.com/TheAlgorithms/Python/blob/a9f73e318cddf43769083614a3e1f9dab1ec50fc/data_structures/linked_list/skip_list.py
```

pydatastructs/linear_data_structures/linked_lists.py
```diff
@@ -722,3 +724,213 @@ def extract(self, index):
...
+
+    """
+
+    __slots__ = ['size', 'level', 'p', 'max_level']
+
+    def __new__(cls, p=0.5, max_level=16):
```
czgdp1807:
```suggestion
    def __new__(cls, p, levels):
```

pydatastructs/utils/misc_util.py
```diff
@@ -170,6 +171,33 @@ def __new__(cls, data=None, links=['next'], addrs=[None]):
     def __str__(self):
         return str(self.data)
 
+class SkipListNode(Node):
```
czgdp1807: Why can't we simply use `LinkedListNode` with number of `links` equal to the number of levels in `SkipList`?

HarsheetKakar: `links=['next', 'forward']`, `addrs=[None,[]]` is this what you are suggesting?

HarsheetKakar: I would have to look into SinglyLinkedList implementation for that, can we do it later, after RedBlackTree is complete?

pydatastructs/linear_data_structures/linked_lists.py
HarsheetKakar: max_level is how much skip list can extend vertically it is different from how many actual levels are there

czgdp1807:
This [lecture](https://www.cs.cmu.edu/~ckingsf/bioinfo-lectures/skiplists.pdf) explains skip lists in a much better way than the original paper.
It looks like there isn't a need of methods other than `insert`, `delete` and `search` for now. `insert` and `delete` can use `search` as the helper method.
In addition, a minor improvement is needed `LinkedListNode`. It lacks, `key` attribute. Will make a PR to modify it.

pydatastructs/linear_data_structures/linked_lists.py
HarsheetKakar: In the implementations I have seen, there is always some form of default value for frequency of elements in i+1 level. 0.5 was used in MIT lecture.

czgdp1807: `p=0.5` is mostly used for perfect skip lists. Let's leave it upto to the programmer to decide what value they want to use.

HarsheetKakar: so no default value for p?

HarsheetKakar:
> This [lecture](https://www.cs.cmu.edu/~ckingsf/bioinfo-lectures/skiplists.pdf) explains skip lists in a much better way than the original paper.
> It looks like there isn't a need of methods other than `insert`, `delete` and `search` for now. `insert` and `delete` can use `search` as the helper method.
> In addition, a minor improvement is needed `LinkedListNode`. It lacks, `key` attribute. Will make a PR to modify it.

main methods are insert delete and find only rest are helpers.

pydatastructs/linear_data_structures/linked_lists.py
czgdp1807: > max_level is how much skip list can extend vertically it is different from how many actual levels are there

[This](https://epaperpress.com/sortsearch/download/skiplist.pdf) defines MaxLevel as L(N) where N is the upper bound on number of elements which we don't know at the start. Can we do something like providing an option to the user to provide the maximum number of elements that their skip list can have and calculate max levels on the basis of that.

czgdp1807:
> main methods are insert delete and find only rest are helpers.

Yes, the helper methods are too many. Can we keep these methods self contained?

pydatastructs/linear_data_structures/linked_lists.py
HarsheetKakar: we can do that for sure

HarsheetKakar:
> > main methods are insert delete and find only rest are helpers.
> 
> Yes, the helper methods are too many. Can we keep these methods self contained?

there are 2 helper functions `_random_level` and `_locate_node`, I can remove `_random_level` maybe because it is referenced only in insert function but `_locate_node` is referenced in all 3 main functions, I will result in duplication of code.

pydatastructs/linear_data_structures/linked_lists.py
HarsheetKakar: https://github.com/geertj/pyskiplist/blob/c5f94cf135d42bb277255150d3f570ed807468b2/pyskiplist/skiplist.py#L115 in here also they have given max_level by default. I guess the main reason is if max level increases more than log(N) where N is the number of elements, the cost will become worse than binary tree. Increasing the max_level will also mean increasing duplication and since we dont know before hand how many elements are supposed to be there, its better that programmer himself assigns the value.

HarsheetKakar: Other thing we can do is that if the max_level is given then take that as max_level else if list of elements is given then max_level= log(N) else take the default value of 16.

HarsheetKakar: > > max_level is how much skip list can extend vertically it is different from how many actual levels are there
> 
> [This](https://epaperpress.com/sortsearch/download/skiplist.pdf) defines MaxLevel as L(N) where N is the upper bound on number of elements which we don't know at the start. Can we do something like providing an option to the user to provide the maximum number of elements that their skip list can have and calculate max levels on the basis of that.

Shall I add a parameter for expected number of elements?

pydatastructs/linear_data_structures/linked_lists.py
```diff
@@ -736,3 +738,213 @@ def extract(self, index):
...
+        obj.head = SkipListNode("root", None)
+        obj.level = 0
+        obj.p = p
+        obj.max_level = max_level
+        obj.size = 0
+        return obj
+
+    def __str__(self):
+        """
+        :return: Visual representation of SkipList
```
czgdp1807: Please use numpy doc string guide.

HarsheetKakar: ok will do that.

HarsheetKakar: Sir please take a look at tests also, and check if its fine with you.

czgdp1807:
Some points to consider,

1. Can we make an array of head pointers of each linked list with index of any head denoting the linked list to which it belongs? I do not which representation is used right now.
2. To print the skip list, can we print a matrix, with number of rows as number of levels and the number of columns as the size of the linked list at the base. The topmost linked list should have the least number of nodes, and the last linked list should have all nodes. We can fill the list column wise, if an element is not present in a list a simple flag value like, ' ' this can be printed.

HarsheetKakar:
> Some points to consider,
> 
> 1. Can we make an array of head pointers of each linked list with index of any head denoting the linked list to which it belongs? I do not which representation is used right now.

It is actually using list vertically. Imagine the lowest nodes to be connected in a linked list and each of them having a list which can at max go to `max_level`.

> 2. To print the skip list, can we print a matrix, with number of rows as number of levels and the number of columns as the size of the linked list at the base. The topmost linked list should have the least number of nodes, and the last linked list should have all nodes. We can fill the list column wise, if an element is not present in a list a simple flag value like, ' ' this can be printed.

will start working on it

HarsheetKakar:
the implementation is not efficient, will have to create it again.

PR #157, opened 2020-03-15
Added implementation for Splay Trees.

Vanshika266:
Added code for splay trees with public APIs insert, delete, splay. The work is in reference with issue #109

PR #176, opened 2020-03-19
[WIP] Added base class SelfBalancingTree for Red Black Tree

HarsheetKakar:

Please check if the made changes are advisable or should I just extend RedBlackTree from AVLTree.

Added SelfBalancingTree which consists rotations.

### DIFF
--- a/pydatastructs/trees/binary_trees.py
+++ b/pydatastructs/trees/binary_trees.py
@@ -552,7 +552,7 @@ class BinarySearchTree(BinaryTree):
         """
         return getattr(self, "_lca_"+str(algorithm))(j, k)
 
-class SelfBalanceingTree(BinarySearchTree):
+class SelfBalancingTree(BinarySearchTree):
     """
     Represents Base class for all rotation based balancing trees like AVL tree, Red Black tree, Splay Tree.
     """
--- a/pydatastructs/trees/binary_trees.py
+++ b/pydatastructs/trees/binary_trees.py
@@ -552,7 +552,73 @@ class BinarySearchTree(BinaryTree):
         """
         return getattr(self, "_lca_"+str(algorithm))(j, k)
 
-class AVLTree(BinarySearchTree):
+class SelfBalanceingTree(BinarySearchTree):
+    """
+    Represents Base class for all rotation based balancing trees like AVL tree, Red Black tree, Splay Tree.
+    """
+    def _right_rotate(self, j, k):
+        y = self.tree[k].right
+        if y is not None:
+            self.tree[y].parent = j
+        self.tree[j].left = y
+        self.tree[k].parent = self.tree[j].parent
+        if self.tree[k].parent is not None:
+            self.tree[self.tree[k].parent].left = k
+        self.tree[j].parent = k
+        self.tree[k].right = j
+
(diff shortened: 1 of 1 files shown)


czgdp1807:
Looks like a nice idea. Tests are also passing without modifications. The lines not covered in diff were not originally covered so not a concern of this PR. It looks like ready for merging.

HarsheetKakar:
ok so merge this one and I will try to implement RBTree by tomorrow.

pydatastructs/trees/binary_trees.py
```diff
@@ -552,29 +552,10 @@ def lowest_common_ancestor(self, j, k, algorithm=1):
         """
         return getattr(self, "_lca_"+str(algorithm))(j, k)
 
-class AVLTree(BinarySearchTree):
+class SelfBalancingTree(BinarySearchTree):
```
czgdp1807:
```suggestion
class SelfBalancingBinaryTree(BinarySearchTree):
```

czgdp1807:
Are the rotation methods in `_left_rotate` generic enough? For example in https://github.com/codezonediitj/pydatastructs/pull/157 the rotations in `SplayTree` directly re-use the code in `AVLTree` . So, can we directly reuse the rotation methods in `SelfBalancingBinaryTree` in `SplayTree`. I observed that, `AVLTree` still have to do some stuff in their own rotation methods.

HarsheetKakar:
> Are the rotation methods in `_left_rotate` generic enough? For example in #157 the rotations in `SplayTree` directly re-use the code in `AVLTree` . So, can we directly reuse the rotation methods in `SelfBalancingBinaryTree` in `SplayTree`. I observed that, `AVLTree` still have to do some stuff in their own rotation methods.

in AVLTree its just recalculating the height, which is necessary for balancing in AVL but in case of other trees eg. RedBlackTree we dont actually care about the absolute height. Rotations are not part of how a tree is built or what data its nodes hold, it just moves pointers that is why I dont think it will break the code.

HarsheetKakar:
Any way the code is actually cut from the original functions in AVL tree, so even if Splay Tree is extending from AVL tree there shouldn't be any problem.

czgdp1807:
Yes, but what if we subclass `SelfBalancingBinaryTree` in `SplayTree`? Will there be any affect on `SplayTree` in that case?

HarsheetKakar:
> Yes, but what if we subclass `SelfBalancingBinaryTree` in `SplayTree`? Will there be any affect on `SplayTree` in that case?

IMO since they arnt doing anything with the height of tree then the code should be same, but I might have to look into SplayTree to be absolutely sure.

HarsheetKakar:
Just looked into it and I think it wont be a problem if we replace AVLTree with SelfBalancingBinaryTree since the rotations are still the same.

#### Reference:
https://www.youtube.com/watch?v=IBY4NtxmGg8

czgdp1807:
Currently the tests are not failing and if in future there is bug then we may need to restore the changes. Merging for now.

HarsheetKakar:
I am sorry, there is a major bug in SelfBalancingTree class, I am opening a PR to revert changes.

Issue #101, opened 2020-02-26
Add m-ary trees


czgdp1807:
#### Description of the problem

Adding m-ary trees will be a nice idea. May be a array based(allows printing the trees) or pointer(better space utilisation) or both would be better.

vibhu18116:
Hi, can I start working on this issue?

czgdp1807:
Hi, @vibhu18116 Please start by designing an API for m-ary trees. Please let us know if you face any problem/queries.

vibhu18116:
Thank you for allowing me to work on this.

I believe the API design would closely follow that of binary trees. Hence these are a few APIs I think should form a basis:

- ` __new__(cls, m_value = 2, array_flag = 0, key=None, root_data=None, comp=None)`

This would initialise a new m-ary tree. m_value will be used to determine the maximum value of m (the number of children of a node), default to 2 (binary tree in that case). array_flag can be set to indicate that an array implementation is required and it can remain 0 for pointer implementation.

Rest of the parameters will be same as binary tree implementation.

- `insert(self, key, data)`
- `delete(self, key, **kwargs)`
- `search(self, key, **kwargs)`
- `__str__(self)`

The above four will behave in the same way as in the binary tree. Implementation of the `__str__` function will vary depending upon if the tree is implemented through array or pointers.

- `height(self)`
Will return the height of the tree.

- `width(self)`
Will return the width of the tree.

- `convert_to_binary(self, array_flag = 0)`
This will convert an m-ary tree to binary tree and return the pointer to root if array_flag is not set and the array if array_flag is set.

Kindly check if these APIs work.

czgdp1807:
> `__new__(cls, m_value = 2, array_flag = 0, key=None, root_data=None, comp=None)`

IMO, using only array based implementation will be better here as pointers are somewhat unsafe, so avoid them where ever possible. Instead of `m_value`, a better name would be `max_children`. [This](https://en.wikipedia.org/wiki/M-ary_tree#Arrays) suggests a nice way to implement using arrays. 

 Can you please explain the use of the values returned by `height` and `width` methods?

 > `convert_to_binary(self, array_flag = 0)`

IMO, just `to_binary_tree(self)` would better. 

You may refer, https://github.com/codezonediitj/pydatastructs/blob/master/pydatastructs/trees/binary_trees.py while implementing the above APIs.

czgdp1807:
[...]

PR #179, opened 2020-03-19
Added code for Insertion in m-ary trees

Saptashrungi:

### DIFF
--- a/pydatastructs/utils/misc_util.py
+++ b/pydatastructs/utils/misc_util.py
@@ -299,14 +299,25 @@ class Set(object):
         obj.parent, obj.size = [None]*2
         return obj
 
-class BTreeNode(object):
+'''class BTreeNode(object):
 
-    __slots__ = ['keys', 'degree', 'parent', 'children', 'data', 'is_root', 'no_of_keys', 'search', 'traverse']
+    __slots__ = ['keys', 'degree', 'children', 'data', 'is_root', 'no_of_keys', 'search', 'traverse']
 
     def __new__(cls, key, data, degree, children):
         obj = Node.__new__(cls)
         obj.data, obj.key = data, key
         obj.degree = degree
+        obj.no_of_keys = no_of_keys
         obj.children = [] or children
         obj.is_root = False
         return obj
+
+    def searchNode(node):
+        i = 0
+        while i<no_of_keys and node>keys[i]:
+            i = i + 1
+            if keys[i] == node:
+                return self
+            if children == []
+                return None
+            return children[i].searchNode(node)'''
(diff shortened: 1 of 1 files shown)


pydatastructs/trees/m_ary_trees.py
```diff
@@ -77,7 +77,13 @@ def insert(self, key, data):
...
         None
         """
-        raise NotImplementedError("This is an abstract method.")
+        temp = self.search(key)
+        if temp.is_root is True and temp.key is None:
+            temp.key = key
+            temp.data = data
+        if temp is not None:
+            self.tree[temp].data = data
+
```
czgdp1807: Please see https://github.com/codezonediitj/pydatastructs/issues/101#issuecomment-595095244

pydatastructs/utils/misc_util.py
```diff
@@ -297,3 +298,26 @@ def __new__(cls, key, data):
...
+    __slots__ = ['keys', 'degree', 'children', 'data', 'is_root', 'no_of_keys', 'search', 'traverse']
+
+    def __new__(cls, key, data, degree, children):
+        obj = Node.__new__(cls)
+        obj.data, obj.key = data, key
+        obj.degree = degree
+        obj.no_of_keys = no_of_keys
+        obj.children = [] or children
+        obj.is_root = False
+        return obj
```
czgdp1807: https://github.com/codezonediitj/pydatastructs/blob/692624605aeb8a47c33f82e4838f8a01b8af5d3d/pydatastructs/utils/misc_util.py#L136-L143

No of keys can be figured out from no of children.
Is degree the same as order in Knuth's definition.

pydatastructs/trees/m_ary_trees.py
Saptashrungi: So we need only search function for m-ary?

czgdp1807:
@vibhu18116 is already working on `MArySearchTree`. In addition, APIs for B-Tree depend on `MArySearchTree`. So, let's wait for @vibhu18116 to complete `MArySearchTree` first. 
Closing.

pydatastructs/trees/m_ary_trees.py
czgdp1807: `MAryTree` is abstract. `MArySearchTree` is concrete in the way insertion and search is done, so both of these method will be defined in that class.

pydatastructs/utils/misc_util.py
Saptashrungi: Degree represents the lower bound on the number of children a node in the B Tree can have (except for the root). i.e the minimum number of children possible. Whereas the Order represents the upper bound on the number of children. ie. the maximum number possible.

czgdp1807: > Every non-leaf node (except root) has at least ⌈m/2⌉ child nodes.

Can degree be computed from order?

PR #189, opened 2020-03-22
SelfBalancingTree updated

HarsheetKakar:

root_idx was not updating after #176


HarsheetKakar:
@czgdp1807 please check it.

czgdp1807:
It isn't really making any big difference for the current code. So merging.

czgdp1807:
The idea of introducing `SelfBalancingTree` is fine, so we will not be reverting that in any case. All the rotation methods should reside under that class.

PR #207, opened 2020-03-24
brick_sort implemented

### LINKED ISSUE #129: Parallel Brick Sort
I would like to propose an odd-even transposition sort or brick sort. Odd-Even Transposition Sort is based on the Bubble Sort technique. It compares two adjacent numbers and switches them if the first number is greater than the second number to get an ascending order list. The opposite case applies to a descending order series.

Please assign me this issue.

HarsheetKakar:

#129

### DIFF
--- a/pydatastructs/linear_data_structures/algorithms.py
+++ b/pydatastructs/linear_data_structures/algorithms.py
@@ -137,7 +137,7 @@ def brick_sort(array, **kwargs):
     >>> from pydatastructs import OneDimensionalArray, brick_sort
     >>> arr = OneDimensionalArray(int,[3, 2, 1])
     >>> brick_sort(arr)
-    >>> print(list(arr))
+    >>> [arr[0], arr[1], arr[2]]
     [1, 2, 3]
     >>> brick_sort(arr, comp=lambda u, v: u > v)
     >>> print(list(arr))
--- a/pydatastructs/linear_data_structures/algorithms.py
+++ b/pydatastructs/linear_data_structures/algorithms.py
@@ -147,22 +147,17 @@ def brick_sort(array, **kwargs):
     ==========
     .. [1] https://www.geeksforgeeks.org/odd-even-sort-brick-sort/
     """
-
     start = kwargs.get('start', 0)
     end = kwargs.get('end', len(array) - 1)
     comp = kwargs.get("comp", lambda u, v: u <= v)
 
     is_sorted = False
-
     while is_sorted is False:
-
         is_sorted = True
-
         for i in range(start+1, end, 2):
             if comp(array[i+1], array[i]):
                 array[i], array[i+1] = array[i+1], array[i]
                 is_sorted = False
-
         for i in range(start, end, 2):
             if comp(array[i+1], array[i]):
                 array[i], array[i+1] = array[i+1], array[i]
--- a/pydatastructs/linear_data_structures/algorithms.py
+++ b/pydatastructs/linear_data_structures/algorithms.py
@@ -148,7 +148,7 @@ def brick_sort(array, **kwargs):
     .. [1] https://www.geeksforgeeks.org/odd-even-sort-brick-sort/
     """
     start = kwargs.get('start', 0)
-    end = kwargs.get('end', len(array) - 1)
+    end = kwargs.get('end', array._last_pos_filled if _check_type(array, DynamicArray) else len(array) - 1)
     comp = kwargs.get("comp", lambda u, v: u <= v)
 
     is_sorted = False
@@ -162,3 +162,6 @@ def brick_sort(array, **kwargs):
             if comp(array[i+1], array[i]):
                 array[i], array[i+1] = array[i+1], array[i]
                 is_sorted = False
+
+    if _check_type(array, DynamicArray):
+        array._modify(force=True)
--- a/pydatastructs/linear_data_structures/algorithms.py
+++ b/pydatastructs/linear_data_structures/algorithms.py
@@ -148,18 +148,18 @@ def brick_sort(array, **kwargs):
     .. [1] https://www.geeksforgeeks.org/odd-even-sort-brick-sort/
     """
     start = kwargs.get('start', 0)
-    end = kwargs.get('end', array._last_pos_filled if _check_type(array, DynamicArray) else len(array) - 1)
+    end = kwargs.get('end', len(array) - 1)
     comp = kwargs.get("comp", lambda u, v: u <= v)
 
     is_sorted = False
     while is_sorted is False:
         is_sorted = True
         for i in range(start+1, end, 2):
-            if comp(array[i+1], array[i]):
+            if _comp(array[i+1], array[i], comp):
                 array[i], array[i+1] = array[i+1], array[i]
                 is_sorted = False
         for i in range(start, end, 2):
-            if comp(array[i+1], array[i]):
+            if _comp(array[i+1], array[i], comp):
                 array[i], array[i+1] = array[i+1], array[i]
                 is_sorted = False
 
--- a/pydatastructs/linear_data_structures/algorithms.py
+++ b/pydatastructs/linear_data_structures/algorithms.py
@@ -5,7 +5,8 @@ from concurrent.futures import ThreadPoolExecutor
 from math import log, floor
 
 __all__ = [
-    'merge_sort_parallel'
+    'merge_sort_parallel',
+    'brick_sort'
 ]
 
 def _merge(array, sl, el, sr, er, end, comp):
@@ -104,3 +105,65 @@ def merge_sort_parallel(array, num_threads, **kwargs):
 
     if _check_type(array, DynamicArray):
         array._modify(force=True)
+
+def brick_sort(array: OneDimensionalArray, **kwargs):
+    """
+    Implements Brick Sort / Odd Even sorting algorithm
+
+    Parameters
+    ==========
+
+    array: Array
+        The array which is to be sorted.
+    start: int
+        The starting index of the portion
+        which is to be sorted.
+        Optional, by default 0
+    end: int
+        The ending index of the portion which
+        is to be sorted.
+        Optional, by default the index
+        of the last position filled.
+    comp: lambda/function
+        The comparator which is to be used
+        for sorting. If the function returns
+        False then only swapping is performed.
+        Optional, by default, less than or
+        equal to is used for comparing two
+        values.
+
+    Examples
+    ========
+    >>> from pydatastructs import OneDimensionalArray, brick_sort
+    >>> arr = OneDimensionalArray(int,[3, 2, 1])
+    >>> brick_sort(arr)
+    >>> print(list(arr))
+    [1, 2, 3]
+    >>> brick_sort(arr, comp=lambda u, v: u > v)
+    >>> print(list(arr))
+    [3, 2, 1]
+
+    References
+    ==========
+    .. [1] https://www.geeksforgeeks.org/odd-even-sort-brick-sort/
+    """
+
+    start = kwargs.get('start', 0)
+    end = kwargs.get('end', len(array) - 1)
+    comp = kwargs.get("comp", lambda u, v: u <= v)
+
+    is_sorted = False
+
+    while is_sorted is False:
+
+        is_sorted = True
+
+        for i in range(start+1, end, 2):
+            if comp(array[i+1], array[i]):
+                array[i], array[i+1] = array[i+1], array[i]
+                is_sorted = False
+
+        for i in range(start, end, 2):
+            if comp(array[i+1], array[i]):
+                array[i], array[i+1] = array[i+1], array[i]
+                is_sorted = False
--- a/pydatastructs/linear_data_structures/algorithms.py
+++ b/pydatastructs/linear_data_structures/algorithms.py
@@ -140,7 +140,7 @@ def brick_sort(array, **kwargs):
     >>> [arr[0], arr[1], arr[2]]
     [1, 2, 3]
     >>> brick_sort(arr, comp=lambda u, v: u > v)
-    >>> print(list(arr))
+    >>> [arr[0], arr[1], arr[2]]
     [3, 2, 1]
 
     References
--- a/pydatastructs/linear_data_structures/algorithms.py
+++ b/pydatastructs/linear_data_structures/algorithms.py
@@ -106,7 +106,7 @@ def merge_sort_parallel(array, num_threads, **kwargs):
     if _check_type(array, DynamicArray):
         array._modify(force=True)
 
-def brick_sort(array: OneDimensionalArray, **kwargs):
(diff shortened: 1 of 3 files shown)


pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -104,3 +105,65 @@ def merge_sort_parallel(array, num_threads, **kwargs):
 
     if _check_type(array, DynamicArray):
         array._modify(force=True)
+
+def brick_sort(array: OneDimensionalArray, **kwargs):
```
czgdp1807:
```suggestion
def brick_sort(array, **kwargs):
```

pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -104,3 +105,65 @@ def merge_sort_parallel(array, num_threads, **kwargs):
...
+
+        for i in range(start+1, end, 2):
+            if comp(array[i+1], array[i]):
+                array[i], array[i+1] = array[i+1], array[i]
+                is_sorted = False
+
+        for i in range(start, end, 2):
+            if comp(array[i+1], array[i]):
+                array[i], array[i+1] = array[i+1], array[i]
+                is_sorted = False
```
czgdp1807:
```suggestion
    start = kwargs.get('start', 0)
    end = kwargs.get('end', len(array) - 1)
    comp = kwargs.get("comp", lambda u, v: u <= v)

    is_sorted = False
    while is_sorted is False:
        is_sorted = True
        for i in range(start+1, end, 2):
            if comp(array[i+1], array[i]):
                array[i], array[i+1] = array[i+1], array[i]
                is_sorted = False
        for i in range(start, end, 2):
            if comp(array[i+1], array[i]):
                array[i], array[i+1] = array[i+1], array[i]
                is_sorted = False
```

pydatastructs/linear_data_structures/tests/test_algorithms.py
```diff
@@ -31,3 +31,17 @@ def test_merge_sort_parallel():
...
+    n = 10
+    arr = OneDimensionalArray(int, [random.randint(1,1000) for _ in range(n)])
+    b = sorted(arr)
+    brick_sort(arr)
+    assert b == list(arr)
+
+    arr = OneDimensionalArray(int, [random.randint(1,1000) for _ in range(n)])
+    brick_sort(arr, comp=lambda u, v: u>v)
+    b = list(reversed(sorted(arr)))
+    assert b == list(arr)
```
czgdp1807: Please test your code for `DynamicOneDimensionalArray` as well. It may require to make some changes to your code to make things work.

HarsheetKakar: ok will do that

pydatastructs/linear_data_structures/tests/test_algorithms.py
```diff
@@ -31,3 +31,26 @@ def test_merge_sort_parallel():
                     709, 910]
     merge_sort_parallel(arr, 5, start=2, end=5)
     assert arr._data == expected_arr
+
+def test_brick_sort():
```
czgdp1807: May be we can use the test case in `test_merge_sort` for all the sorting algorithms. Something like this is also done with `minimum_spanning_tree`. Refer, https://github.com/codezonediitj/pydatastructs/blob/6605b9761a920632a69e34547765a2c2b78cbb13/pydatastructs/graphs/tests/test_algorithms.py#L154-L179

HarsheetKakar: ok so one common Array is sorted by multiple sorting algorithms, is that what you mean?

czgdp1807: Yes.

HarsheetKakar: shall I implement it in this PR itself or should I make another one?

czgdp1807: In this PR itself.

HarsheetKakar: Oh I get ur point, delete leaves None in the place where the value was originally, I might have to change many things in the code. Thanks it was very valuable input :)

czgdp1807: Probably, just `None`s have to hr ignored while comparing.

HarsheetKakar: One question. if [1,2,3,4,None,None,5,6] is the array and I sort from 3 to 6 should the None be in its original position or should it move to the end of the array.

HarsheetKakar: > Probably, just `None`s have to be ignored while comparing. `None`s can cause problems in many places. While porting to the code to C++, `NULL` can cause problems that are currently caused by `None`.

cant we overload the operators?

HarsheetKakar: another solution is to change delete to remove the element and reconfigure the array(thats how list is doing it)

czgdp1807: No, that shouldn't be done. The difference between `list` and DODA is the time complexity of deletion. `list` does the deletion in O(n) time but `DODA` does the same in amortized O(1) time. `None` is used as flag value in `DODA` that the element is deleted from a particular case.

HarsheetKakar: understood

HarsheetKakar: > One question. if [1,2,3,4,None,None,5,6] is the array and I sort from 3 to 6 should the None be in its original position or should it move to the end of the array.

and how should this be done?

czgdp1807: The following is what `merge_sort_parallel` does,
```python
>>> from pydatastructs import merge_sort_parallel
>>> from pydatastructs import DynamicOneDimensionalArray
>>> DODA = DynamicOneDimensionalArray
>>> arr = DODA(int, 0)
>>> arr.append(1)
>>> arr.append(2)
>>> arr.append(3)
>>> arr.append(4)
>>> arr.append(9)
>>> arr.append(10)
>>> arr.append(5)
>>> arr.append(6)
>>> arr._data
[1, 2, 3, 4, 9, 10, 5, 6, None, None, None, None, None, None, None]
>>> arr.delete(4)
>>> arr.delete(5)
>>> arr._data
[1, 2, 3, 4, None, None, 5, 6, None, None, None, None, None, None, None]
>>> merge_sort_parallel(arr, 1, start=2, end=7)
>>> arr._data
[1, 2, 3, 4, 5, 6, None, None, None, None, None, None, None]
```
`brick_sort` can do the same.

HarsheetKakar: merge_sort_parallel(arr, 1, start=2, end=5)
>>> arr._data
[1, 2, 3, 4, 5, 6, None, None, None, None, None, None, None]

as I was sorting from just 2 to 5 the array should have been [3,4,None,None,5] but the None values are moved to the end, so is that what you want to do or would it be better if the elements which are not included in the sort retain their original positions?

czgdp1807: Well, 2 to 5 includes `None`s, so change their positions should be expected.

HarsheetKakar: but that changes the positions of other elements as well, I can do it if u want but idk if users should expect other indexes to change if they sort a small segment.

HarsheetKakar: lets raise this issue some other time for some other PR, I will handle None's as you have done in merge sort

czgdp1807: I am looking into this issue for now. If it's fixed with minimal changes then we can go for it.

pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -104,3 +105,63 @@ def merge_sort_parallel(array, num_threads, **kwargs):
...
+
+    is_sorted = False
+    while is_sorted is False:
+        is_sorted = True
+        for i in range(start+1, end, 2):
+            if comp(array[i+1], array[i]):
+                array[i], array[i+1] = array[i+1], array[i]
+                is_sorted = False
+        for i in range(start, end, 2):
+            if comp(array[i+1], array[i]):
```
czgdp1807:
```suggestion
            if _comp(array[i+1], array[i], comp):
```

pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -104,3 +105,63 @@ def merge_sort_parallel(array, num_threads, **kwargs):
...
+    """
+    start = kwargs.get('start', 0)
+    end = kwargs.get('end', array._last_pos_filled if _check_type(array, DynamicArray) else len(array) - 1)
+    comp = kwargs.get("comp", lambda u, v: u <= v)
+
+    is_sorted = False
+    while is_sorted is False:
+        is_sorted = True
+        for i in range(start+1, end, 2):
+            if comp(array[i+1], array[i]):
```
czgdp1807:
```suggestion
            if _comp(array[i+1], array[i], comp):
```

pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -104,3 +105,63 @@ def merge_sort_parallel(array, num_threads, **kwargs):
...
+    >>> brick_sort(arr, comp=lambda u, v: u > v)
+    >>> print(list(arr))
+    [3, 2, 1]
+
+    References
+    ==========
+    .. [1] https://www.geeksforgeeks.org/odd-even-sort-brick-sort/
+    """
+    start = kwargs.get('start', 0)
+    end = kwargs.get('end', array._last_pos_filled if _check_type(array, DynamicArray) else len(array) - 1)
```
czgdp1807:
```suggestion
    end = kwargs.get('end', len(array) - 1)
```

pydatastructs/linear_data_structures/tests/test_algorithms.py
czgdp1807: Resolving.

pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -104,3 +105,63 @@ def merge_sort_parallel(array, num_threads, **kwargs):
...
+        Optional, by default, less than or
+        equal to is used for comparing two
+        values.
+
+    Examples
+    ========
+    >>> from pydatastructs import OneDimensionalArray, brick_sort
+    >>> arr = OneDimensionalArray(int,[3, 2, 1])
+    >>> brick_sort(arr)
+    >>> print(list(arr))
```
czgdp1807:
```suggestion
    >>> [arr[0], arr[1], arr[2]]
```

pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -104,3 +105,63 @@ def merge_sort_parallel(array, num_threads, **kwargs):
...
+    [1, 2, 3]
+    >>> brick_sort(arr, comp=lambda u, v: u > v)
+    >>> print(list(arr))
```
czgdp1807:
```suggestion
    >>> [arr[0], arr[1], arr[2]]
```

HarsheetKakar:
so is it ready to merge now? @czgdp1807

czgdp1807:
> so is it ready to merge now? @czgdp1807

Tests should be updated. Common tests should be used for all sorting algorithms. The function, `test_merge_sort_parallel` can be renamed to `_test_common_sort(sort_func)` and `test_merge_sort_parallel` and `test_brick_sort` should call them from inside, like,
```python
def test_merge_sort_parallel():
    _test_common_sort(merge_sort_parallel)
```

HarsheetKakar:
> > so is it ready to merge now? @czgdp1807
> 
> Tests should be updated. Common tests should be used for all sorting algorithms. The function, `test_merge_sort_parallel` can be renamed to `_test_common_sort(sort_func)` and `test_merge_sort_parallel` and `test_brick_sort` should call them from inside, like,
> 
> ```python
> def test_merge_sort_parallel():
>     _test_common_sort(merge_sort_parallel)
> ```

sure sounds good, and while we are at it I also feel that we should put the algorithms under their own module.

czgdp1807:
Or testing function for parallel and non parallel algorithms can be separated(array(s) used in testing should be same, only `num_threads` needs attention). I hope you are getting my point.

czgdp1807:
> while we are at it I also feel that we should put the algorithms under their own module.

Well, every `algorithm.py` file under different modules contains algorithms for the data structure represented by that module. This helps in keeping things specific for each module. Every algorithm is under global namespace so not a big issue.

HarsheetKakar:
> Or testing function for parallel and non parallel algorithms can be separated(array(s) used in testing should be same, only `num_threads` needs attention). I hope you are getting my point.

I get it, and I have felt the same while I was working on it... so this is the proposal:
>pydatastructs
>>algorithms
>>>sequential
>>>>tests

>>>concurrent
>>>>tests

czgdp1807:
> I get it, and I have felt the same while I was working on it... so this is the proposal:

Every parallel algorithm has `_parallel` appended to it, so putting it under different modules. Though for current classes, may not work with parallel algorithms through processes because sharing user defined objects among processes is not possible with python. A `_to_python_object` method can be introduced in every class to support that. Opening a separate issue for this.

pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -104,3 +105,63 @@ def merge_sort_parallel(array, num_threads, **kwargs):
...
+    while is_sorted is False:
+        is_sorted = True
+        for i in range(start+1, end, 2):
+            if _comp(array[i+1], array[i], comp):
+                array[i], array[i+1] = array[i+1], array[i]
+                is_sorted = False
+        for i in range(start, end, 2):
+            if _comp(array[i+1], array[i], comp):
+                array[i], array[i+1] = array[i+1], array[i]
+                is_sorted = False
```
HarsheetKakar: It says _comp not found.
`Fail: NameError: name '_comp' is not defined`

czgdp1807: merge master to your branch `bricksort`, push the changes here.

HarsheetKakar: I think definition for _comp was accidentally removed @czgdp1807

HarsheetKakar: Thanks it worked

HarsheetKakar:
> Or testing function for parallel and non parallel algorithms can be separated(array(s) used in testing should be same, only `num_threads` needs attention). I hope you are getting my point.

take a look at it please.

PR #208, opened 2020-03-24
Added str methods for ODA and DODA

HarsheetKakar:

### DIFF
--- a/pydatastructs/linear_data_structures/arrays.py
+++ b/pydatastructs/linear_data_structures/arrays.py
@@ -265,6 +265,10 @@ class DynamicOneDimensionalArray(DynamicArray, OneDimensionalArray):
                 to_be_printed[i] = str(self._data[i])
         return str(to_be_printed)
 
+    def __reversed__(self):
+        for i in range(self._last_pos_filled, -1, -1):
+            yield self._data[i]
+
 class ArrayForTrees(DynamicOneDimensionalArray):
     """
     Utility dynamic array for storing nodes of a tree.
--- a/pydatastructs/linear_data_structures/arrays.py
+++ b/pydatastructs/linear_data_structures/arrays.py
@@ -130,6 +130,9 @@ class OneDimensionalArray(Array):
     def __len__(self):
         return self._size
 
+    def __str__(self):
+        return str(self._data)
+
 
 class DynamicArray(Array):
     """
@@ -254,6 +257,9 @@ class DynamicOneDimensionalArray(DynamicArray, OneDimensionalArray):
     @property
     def size(self):
         return self._size
+
+    def __str__(self):
+        return str(self._data[:self._last_pos_filled+1])
 
 class ArrayForTrees(DynamicOneDimensionalArray):
     """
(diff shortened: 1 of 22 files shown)


pydatastructs/linear_data_structures/tests/test_arrays.py
```diff
@@ -27,6 +28,7 @@ def test_DynamicOneDimensionalArray():
     A.append(2)
     A.append(3)
     A.append(4)
+    assert str(A) == '[1, 2, 3, 4]'
```
czgdp1807:
```suggestion
```

czgdp1807: Test for https://github.com/codezonediitj/pydatastructs/pull/208/files#diff-df3f437c897b2969b957400929ff64f7R38

HarsheetKakar: I am sorry it is directing me to array.py and I cant understand what I am supposed to do, can you please elaborate? thanks

czgdp1807: What's the output when, `A._data`  attains `[4, None, None]`(in the tests below).

HarsheetKakar: Done. Thanks for pointing out.

pydatastructs/linear_data_structures/arrays.py
```diff
@@ -255,6 +258,9 @@ def delete(self, idx):
     def size(self):
         return self._size
 
+    def __str__(self):
+        return str(self._data[:self._last_pos_filled+1])
```
czgdp1807: Can we shift the following code here, 
https://github.com/codezonediitj/pydatastructs/blob/6605b9761a920632a69e34547765a2c2b78cbb13/pydatastructs/trees/binary_trees.py#L150-L156

And modify the `str` calls wherever, `DODA` is used.

HarsheetKakar: If I use `DODA.__str__()` it wont print the key of each node it would print something like this
`[<pydatastructs.utils.misc_util.TreeNode object at 0x000001F45C54E048>]` so we still have to override the str method in binary_trees.py

czgdp1807:
```python
def __str__(self): 
     to_be_printed = ['' for i in range(self._last_pos_filled + 1)] 
     for i in range(self._last_pos_filled + 1): 
         if self._data[i] is not None:  
             to_be_printed[i] = str(self._data[i])
     return str(to_be_printed) 
```
The above may break tests for trees if their `str` method is removed. We can keep both the methods(`BinaryTree.__str__` and `DODA.__str__`) as it's just about printing things, and they both will look similar.

HarsheetKakar: ok I get it.

HarsheetKakar: here is the thing, if I take str of each element, the output would be `"['1', '2', '3', '4]"` instead of `'[1, 2, 3, 4]'`

czgdp1807: Let it be.

HarsheetKakar: Done.

HarsheetKakar:
Is this ready to move?

PR #216, opened 2020-03-25
Enhancement in sorting algorithms

czgdp1807:
Sorting algorithms now do not disturb positions which are out of scope of `start` and `end`. `None` is treated is infinite value. A custom comparator is introduced for the same.

ping @HarsheetKakar 

#### Other comments
.. [1] https://github.com/codezonediitj/pydatastructs/pull/207#discussion_r397925522

### DIFF
--- a/pydatastructs/linear_data_structures/tests/test_arrays.py
+++ b/pydatastructs/linear_data_structures/tests/test_arrays.py
@@ -32,9 +32,7 @@ def test_DynamicOneDimensionalArray():
     A.delete(15)
     A.delete(-1)
     A.delete(1)
-    print(A._data)
     A.delete(2)
-    print(A._data)
     assert A._data == [4, None, None]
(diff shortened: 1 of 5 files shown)


pydatastructs/linear_data_structures/tests/test_arrays.py
```diff
@@ -32,7 +32,9 @@ def test_DynamicOneDimensionalArray():
     A.delete(15)
     A.delete(-1)
     A.delete(1)
+    print(A._data)
     A.delete(2)
+    print(A._data)
```
czgdp1807:
```suggestion
    A.delete(2)
```

HarsheetKakar:
Thanks it makes work a lot easier

HarsheetKakar:
Can we change None to another Node type so that we can overload its operators which will remove the need for _comp.

PR #222, opened 2020-03-27
Added tests

### LINKED ISSUE #19: Track coverage

The coverage should be near to 100%. However, currently, it stands at 97%. Write tests for the uncovered lines, to increase the coverage.

#### Example of the problem

The uncovered lines of the code are shown in red as can seen in [this report](https://codecov.io/gh/codezonediitj/pydatastructs/src/master/pydatastructs/linear_data_structures/arrays.py). Similarly, look at other files, and write tests so that the red lines are covered.

This issue will never be closed and will be pinned.

HarsheetKakar:

Fixes #19

Added tests for mary node, binomial tree node and added test_m_ary_trees.py

#### Other comments
Added str function for mary node else it would take it as a TreeNode and raise error because there are no left and right attributes

### DIFF
--- a/pydatastructs/utils/misc_util.py
+++ b/pydatastructs/utils/misc_util.py
@@ -148,6 +148,9 @@ class MAryTreeNode(TreeNode):
         for child in children:
             self.children.append(child)
 
+    def __str__(self):
+        return str((self.key, self.data))
+
 
 class LinkedListNode(Node):
     """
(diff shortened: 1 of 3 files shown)


pydatastructs/utils/misc_util.py
```diff
@@ -148,6 +148,9 @@ def add_children(self, *children):
         for child in children:
             self.children.append(child)
 
+    def __str__(self):
+        return str((self.key, self.data))
+
```
czgdp1807:
```suggestion
```
Already present in parent class, `TreeNode`.

HarsheetKakar: gives error because there is no left or right node in MAryTreeNode

PR #240, opened 2020-03-31
added code for both types of random walks

RidhimaKohli:

PR #242, opened 2020-04-01
Added code for RandomWalk

### LINKED ISSUE #65: Add algorithm for DeepWalk

Once we are done adding basic framework for Graphs, we can start thinking of adding DeepWalk algorithm. 
We first need to decide the feasibility of implementation of the above algorithm. 
[1] https://github.com/phanein/deepwalk

Saptashrungi:

"Fixes #65"

I have added code for random walk with restarts. I will write the tests for it, add comments and improve code quality after getting your reviews.

### DIFF
--- a/pydatastructs/graphs/adjacency_list.py
+++ b/pydatastructs/graphs/adjacency_list.py
@@ -20,7 +20,7 @@ class AdjacencyList(Graph):
         for vertex in vertices:
             obj.__setattr__(vertex.name, vertex)
         obj.vertices = [vertex.name for vertex in vertices]
-        obj.nodes = [for vertex in vertices]
+        obj.nodes = [vertex for vertex in vertices]
         obj.edge_weights = dict()
         return obj
 
--- a/pydatastructs/graphs/adjacency_list.py
+++ b/pydatastructs/graphs/adjacency_list.py
@@ -20,6 +20,7 @@ class AdjacencyList(Graph):
         for vertex in vertices:
             obj.__setattr__(vertex.name, vertex)
         obj.vertices = [vertex.name for vertex in vertices]
+        obj.nodes = [for vertex in vertices]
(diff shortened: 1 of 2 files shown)


pydatastructs/graphs/adjacency_list.py
```diff
@@ -20,6 +20,7 @@ def __new__(cls, *vertices):
         for vertex in vertices:
             obj.__setattr__(vertex.name, vertex)
         obj.vertices = [vertex.name for vertex in vertices]
+        obj.nodes = [vertex for vertex in vertices]
```
czgdp1807, line 23: Not needed.

czgdp1807:
Closing due to the reasons similar to https://github.com/codezonediitj/pydatastructs/pull/240#issuecomment-607254446

pydatastructs/graphs/adjacency_list.py
Saptashrungi: I had to access data stored in the node, considering it as index for that node. The vertex name would be a string which can't be used as index to identify node in transition matrix.
The present code is a rough implementation. Presently I have considered only `adjacencylist` here, but it can be applied to matrix as well with some modifications. I have considered the possibility of weighted edges, where `edge.value` is equivalent to weight. I will add tests and improve the code quality later.

czgdp1807: Why do you need the data inside the node/vertex?

Saptashrungi: For now I need something as an index to identify the node. I assumed the data can be used to index.

Saptashrungi: The probability of next node in random walk can be decided based on the weight of the edge or degree according to one of the source you shared. Which one should be implemented.

PR #247, opened 2020-04-05
ArrayDeque added

### LINKED ISSUE #120: Array implementation of doubly ended queue (deque)
I propose the implementation of a doubly ended queue (deque) using arrays. A deque allows insertions and deletions from both the front and the rear end.  The various methods available will be insert front, insert rear, delete front , delete rear in addition to returning the element at front or rear.

Hope this task will be assigned to me

HarsheetKakar:

Fixes #120 

ArrayDeque added
small change in DODA instead of `obj[0]` it is now `obj._data[0]` in `__new__`

### DIFF
--- a/pydatastructs/miscellaneous_data_structures/queue.py
+++ b/pydatastructs/miscellaneous_data_structures/queue.py
@@ -1,11 +1,12 @@
-from pydatastructs.linear_data_structures import DynamicOneDimensionalArray, SinglyLinkedList
+from pydatastructs.linear_data_structures import DynamicOneDimensionalArray, SinglyLinkedList, OneDimensionalArray
 from pydatastructs.utils.misc_util import NoneType, LinkedListNode, _check_type
 from pydatastructs.trees.heaps import BinaryHeap, BinomialHeap
 from copy import deepcopy as dc
 
 __all__ = [
     'Queue',
-    'PriorityQueue'
+    'PriorityQueue',
+    'ArrayDeque'
 ]
 
 class Queue(object):
@@ -68,7 +69,6 @@ class Queue(object):
         raise NotImplementedError(
             "This is an abstract method.")
 
-
 class ArrayQueue(Queue):
 
     __slots__ = ['front']
@@ -125,7 +125,6 @@ class ArrayQueue(Queue):
         for i in range(self.front, self.rear + 1):
             _data.append(self.items._data[i])
         return str(_data)
-
 
 class LinkedListQueue(Queue):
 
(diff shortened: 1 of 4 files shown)


czgdp1807:
Please follow the pattern currently there in `queue.py`. Can we avoid adding new classes?

HarsheetKakar:
> Please follow the pattern currently there in `queue.py`. Can we avoid adding new classes?

add it in the implementation? but the thing is deque is best of stack and queue and is implemented like DODA so I thought it doesnt fit in as another implementation of queue. I can change it if you want.

czgdp1807:
> but the thing is deque is best of stack and queue and is implemented like DODA so I thought it doesnt fit in as another implementation of queue. I can change it if you want.

It should be changed. A doubly ended queue is anyway a queue. You can re-define some methods in the implementation to make things work.

HarsheetKakar:
> > but the thing is deque is best of stack and queue and is implemented like DODA so I thought it doesnt fit in as another implementation of queue. I can change it if you want.
> 
> It should be changed. A doubly ended queue is anyway a queue. You can re-define some methods in the implementation to make things work.

ok

pydatastructs/miscellaneous_data_structures/queue.py
```diff
@@ -364,3 +367,174 @@ def peek(self):
...
+    >>> q.pop()
+    3
+    >>> q.popleft()
+    1
+    >>> q.append(1)
+    >>> q.appendleft(3)
+    >>> [q[i] for i in range(len(q))]
+    [3, 2, 1]
+    """
+
```
czgdp1807, line 430: This doesn't need a documentation.

pydatastructs/miscellaneous_data_structures/queue.py
```diff
@@ -364,3 +367,174 @@ def peek(self):
...
+    def rear(self, val):
+        self._rear = val
+
+    @property
+    def size(self):
+        return self._num
+
+    @property
+    def is_empty(self):
+        return self.size == 0
```
czgdp1807, line 462: May be something using, `self.front` and `self.rear`. What's there in the parent classes? What's the MRO?

pydatastructs/miscellaneous_data_structures/queue.py
```diff
@@ -364,3 +367,174 @@ def peek(self):
...
+
+    @property
+    def is_full(self):
+        return self.size == len(self._data)
```
czgdp1807, line 466: +1

pydatastructs/miscellaneous_data_structures/queue.py
```diff
@@ -364,3 +367,174 @@ def peek(self):
...
+
+    def __getitem__(self, idx):
+        idx = (self.rear + idx) % len(self._data)
+        return self._data[idx]
+
+    def __setitem__(self, idx, val):
+        idx = (self.rear + idx) % len(self._data)
+        self._data[idx] = val
```
czgdp1807, line 474: In queue, elements are accessed from end points so these methods shouldn't be defined.

czgdp1807:
Can we do the following,

```python
class ArrayDeque(ArrayQueue):
   ...
```

pydatastructs/miscellaneous_data_structures/queue.py
HarsheetKakar: tried using front and rear but this worked better since it is circular front can be behind rear in 2 cases if the deque is empty or if it is full and since % operator is used there is no way to find which one is the case

pydatastructs/miscellaneous_data_structures/queue.py
HarsheetKakar: helps a lot in rest of the implementation since I dont have to do modular arithmetic every time I push or pop

czgdp1807: At least they shouldn't be public. You can define helper methods, though 2 line is not much for code indirection.

HarsheetKakar: and these functions are very helpful if u want to change the value of some key in the queue, the user does not have to think about rear and front they can just start from 0 and go up to self.size. It actually makes the code a lot cleaner.

czgdp1807: > the user does not have to think about rear and front

If someone wants to change the value of a key, they could have used some other data structure.

HarsheetKakar: if we dont override it here then if user uses the index it will use DODA's __getitem__ and __setitem__ which will break the logic, and anyway it isnt breaking the already set datastructures and it helps a lot in development also as you can see in append/ appendleft. It makes the code readable, it helps to remove unnecessary bugs and there are only 2 small methods in question.

HarsheetKakar:
> Can we do the following,
> 
> ```python
> class ArrayDeque(ArrayQueue):
>    ...
> ```
Error while trying to put it as ArrayQueue
`Exception has occurred: TypeError
multiple bases have instance lay-out conflict`

pydatastructs/miscellaneous_data_structures/queue.py
HarsheetKakar: and sir it also helps other developers to understand the code.

HarsheetKakar:
@czgdp1807 shall I add LinkedList implementation as well since then both of them would come under Deque abstract class and there would be no inconsistency with the already written code?

czgdp1807:
I think the classes, `ArrayDeque` use `DODA` rather than inheriting and overriding it's methods. You can add `LinkedList` implementation, though again it should be used as a internal data member rather than being inherited.

HarsheetKakar:
> I think the classes, `ArrayDeque` use `DODA` rather than inheriting and overriding it's methods. You can add `LinkedList` implementation, though again it should be used as a internal data member rather than being inherited.

the append and pop method in ArrayDeque is quite different from DODA so if I have to use its properties and not its caveats I have to inherit it. I can also implement it from scratch but it will take more time.

czgdp1807:
Any news?

HarsheetKakar:
> Any news?

sorry forgot about it, give me a couple of days

HarsheetKakar:
> > I think the classes, `ArrayDeque` use `DODA` rather than inheriting and overriding it's methods. You can add `LinkedList` implementation, though again it should be used as a internal data member rather than being inherited.
> 
> the append and pop method in ArrayDeque is quite different from DODA so if I have to use its properties and not its caveats I have to inherit it. I can also implement it from scratch but it will take more time.

and you havent answered this one.

czgdp1807:
Can we do append and pop as follows,
```python
def append(self, x):
    self.queue.append(x)
    self.rear += 1
```

```python
def pop(self, x):
    self.queue.delete(front)
    self.front += 1
    # check corner cases like if it's empty
```

HarsheetKakar:
> Can we do append and pop as follows,
> 
> ```python
> def append(self, x):
>     self.queue.append(x)
>     self.rear += 1
> ```
> 
> ```python
> def pop(self, x):
>     self.queue.delete(front)
>     self.front += 1
>     # check corner cases like if it's empty
> ```

I dont think we can since rear is moved back to 0 if the array is doubled, this helps in keeping the _modify method less complicated.

czgdp1807:
Probably, keeping, `rear` as `self._last_pos_filled` and tracking front by checking when the array will be double should do the the thing.

HarsheetKakar:
> Probably, keeping, `rear` as `self._last_pos_filled` and tracking `front` by checking when the array will be double should do the the thing.

sir but it is doubly ended so that means that rear also would move in the opposite direction, keeping track of both is important... with that said I have forgotten the implementation of most of it and would like to work on some other issue, so if its not a problem with you, we can put `upforgrab` on it... since most of the work is done someone can refactor the code easily

czgdp1807:
Closing in favor of https://github.com/codezonediitj/pydatastructs/pull/321

PR #304, opened 2020-12-07
Replaced the tests of Binary Trees that need manual string matching

### LINKED ISSUE #294: Replace tests of binary tree with in-order and pre-order

As of now, the correctness of operations on a binary tree is determined by verifying the `str(tree)` with the expected string. This makes it quite difficult to check the tests. Instead of using, `str(tree)`, in-order and pre-order should be checked since, using both these traversals a binary tree can be uniquely constructed.

#### Example of the problem

See the tests of RedBlackTree.

[1] https://github.com/codezonediitj/pydatastructs/pull/293

ravisrc:

Closes #294 

Replaced the testing techniques of Binary Trees where initially a manual string has to be fed; to the check of In-order and Pre-order traversals.

#### Other comments

Since a unique tree exists with a set of In-order and Pre-order traversals, this check is comparatively easy to use as there is no need of feeding a string of the nodes, instead of which two sets of keys of nodes are fed.

### DIFF
--- a/pydatastructs/trees/tests/test_binary_trees.py
+++ b/pydatastructs/trees/tests/test_binary_trees.py
@@ -19,6 +19,11 @@ def test_BinarySearchTree():
     b.insert(7, 7)
     b.insert(14, 14)
     b.insert(13, 13)
+    # Explicit check for the __str__ method of Binary Trees Class
+    assert str(b) == \
+    ("[(1, 8, 8, 2), (3, 3, 3, 4), (None, 10, 10, 7), (None, 1, 1, None), "
+    "(5, 6, 6, 6), (None, 4, 4, None), (None, 7, 7, None), (8, 14, 14, None), "
+    "(None, 13, 13, None)]")
 
     trav = BinaryTreeTraversal(b)
     in_order = trav.depth_first_search(order='in_order')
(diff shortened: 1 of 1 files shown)


czgdp1807:
Seems good. Will merge tomorrow.

czgdp1807:
There is [this function](https://codecov.io/gh/codezonediitj/pydatastructs/pull/304/changes) which isn't tested. May be the tests for only this method can be kept along side the new tests.

ravisrc:
> There is [this function](https://codecov.io/gh/codezonediitj/pydatastructs/pull/304/changes) which isn't tested. May be the tests for only this method can be kept along side the new tests.

There since I have replaced all the tests of the binary trees ,and all its derivatives. The `__str__` method is not at all used in the tests except in one position which is [here](https://github.com/codezonediitj/pydatastructs/pull/304/files#diff-6f6c10d84396f704e564796103592332d0c33b857785a46f32126d68a5c8716cL389)

But here comes two issues that the `__str__` method is [redefined](https://github.com/codezonediitj/pydatastructs/blob/master/pydatastructs/trees/binary_trees.py#L748) in the Cartesian tree, which is not at all called in the test_binary_tree. 

So code coverage has been reduced.

ravisrc:
Shall I add that test for the `__str__` for both the basic binary tree and the redefined `__str__` and add a commit to this

czgdp1807:
Yes. Sure

ravisrc:
I have added the tests for the `__str__` methods

PR #316, opened 2021-01-18
potential code refactor, optimize code performance 

tusharnankani:

This PR includes: 

- Remove unused imported modules [Anti-pattern]

- Remove unnecessary `else` / `elif` used after `return` or `raise`
  - `return` statement causes the control flow to be disrupted, making the `else` / `elif` block here unnecessary. This doesn't mean you can not use it, but it is recommended to refactor this for better readability.

- Remove unnecessary use of comprehension.
  - It is unnecessary to use a comprehension just to loop over the iterable and create a `list`/`set`/`dict` out of it. Python has a specialized set of tools for this task: the `list`/`set`/`dict` constructors, which are faster and more readable.

- Changing the default arguments.
  - It is recommended not to use a mutable like list or dictionary as a default value to an argument. Python’s default arguments are evaluated once when the function is defined. Using a mutable default argument and mutating it will mutate that object for all future calls to the function as well.

- Remove `dict()` and use `{}`
  - It is slower to call e.g. dict() than using the empty literal, because the name dict must be looked up in the global scope in case it has been rebound. Same for the other two types like list() and tuple(). 
- Use tuple instead of multiple conditions, simplifying if expression.

### DIFF
--- a/pydatastructs/graphs/adjacency_list.py
+++ b/pydatastructs/graphs/adjacency_list.py
@@ -20,7 +20,7 @@ class AdjacencyList(Graph):
         for vertex in vertices:
             obj.__setattr__(vertex.name, vertex)
         obj.vertices = [vertex.name for vertex in vertices]
-        obj.edge_weights = dict()
+        obj.edge_weights = {}
         return obj
 
     @classmethod
--- a/pydatastructs/graphs/graph.py
+++ b/pydatastructs/graphs/graph.py
@@ -55,14 +55,14 @@ class Graph(object):
             obj = AdjacencyList(*args)
             obj._impl = implementation
             return obj
-        if implementation is 'adjacency_matrix':
+        elif implementation is 'adjacency_matrix':
             from pydatastructs.graphs.adjacency_matrix import AdjacencyMatrix
             obj = AdjacencyMatrix(*args)
             obj._impl = implementation
             return obj
-
-        raise NotImplementedError("%s implementation is not a part "
-                                    "of the library currently."%(implementation))
+        else:
+            raise NotImplementedError("%s implementation is not a part "
+                                        "of the library currently."%(implementation))
 
     def is_adjacent(self, node1, node2):
         """
--- a/pydatastructs/graphs/graph.py
+++ b/pydatastructs/graphs/graph.py
@@ -62,7 +62,7 @@ class Graph(object):
             return obj
         else:
             raise NotImplementedError("%s implementation is not a part "
-                                        "of the library currently."%(implementation))
+                                      "of the library currently."%(implementation))
 
     def is_adjacent(self, node1, node2):
         """
--- a/pydatastructs/graphs/graph.py
+++ b/pydatastructs/graphs/graph.py
@@ -55,14 +55,14 @@ class Graph(object):
             obj = AdjacencyList(*args)
             obj._impl = implementation
             return obj
-        elif implementation is 'adjacency_matrix':
+        if implementation is 'adjacency_matrix':
             from pydatastructs.graphs.adjacency_matrix import AdjacencyMatrix
             obj = AdjacencyMatrix(*args)
             obj._impl = implementation
             return obj
-        else:
-            raise NotImplementedError("%s implementation is not a part "
-                                      "of the library currently."%(implementation))
+
+        raise NotImplementedError("%s implementation is not a part "
+                                    "of the library currently."%(implementation))
 
     def is_adjacent(self, node1, node2):
         """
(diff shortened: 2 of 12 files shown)


pydatastructs/miscellaneous_data_structures/queue.py
```diff
@@ -239,14 +239,14 @@ def __new__(cls, implementation='binary_heap', **kwargs):
...
-        elif implementation == 'binary_heap':
+        if implementation == 'binary_heap':
             return BinaryHeapPriorityQueue(comp)
-        elif implementation == 'binomial_heap':
+        if implementation == 'binomial_heap':
             return BinomialHeapPriorityQueue()
-        else:
-            raise NotImplementedError(
-                "%s implementation is not currently supported "
-                "by priority queue.")
```
czgdp1807, line 249: We need these bunch of `if-elif` for better readability. Please restore the original versions here.

pydatastructs/trees/binary_trees.py
```diff
@@ -515,7 +515,7 @@ def _lca_2(self, j, k):
             else:
                 curr_root = self.tree[curr_root].right
 
-            if curr_root == u or curr_root == v:
```
czgdp1807, line 518: Please restore this.

pydatastructs/trees/heaps.py
```diff
@@ -161,9 +161,8 @@ def insert(self, key, data=None):
             parent = (i - 1)//self.d
             if i == 0 or self._comp(self.heap[parent].key, self.heap[i].key):
                 break
-            else:
-                self._swap(i, parent)
-                i = parent
```
czgdp1807, line 166: `else` needed for better readability.

pydatastructs/utils/misc_util.py
```diff
@@ -436,8 +440,8 @@ def _comp(u, v, tcomp):
     """
     if u is None and v is not None:
         return False
-    elif u is not None and v is None:
+    if u is not None and v is None:
         return True
-    elif u is None and v is None:
```
czgdp1807, line 441: Please restore.

pydatastructs/utils/tests/test_misc_util.py
```diff
@@ -35,7 +35,7 @@ def test_BinomialTreeNode():
 
 def test_MAryTreeNode():
     m = MAryTreeNode(1, 1)
-    m.add_children(*[i for i in range(2,10)])
+    m.add_children(*list(range(2,10)))
```
czgdp1807, line 38:
```suggestion
    m.add_children(*list(range(2, 10)))
```

pydatastructs/graphs/adjacency_list.py
```diff
@@ -20,7 +20,7 @@ def __new__(cls, *vertices):
         for vertex in vertices:
             obj.__setattr__(vertex.name, vertex)
         obj.vertices = [vertex.name for vertex in vertices]
-        obj.edge_weights = dict()
+        obj.edge_weights = {}
```
czgdp1807, line 23: The modification is significantly faster than the original version. However, this introduces ambiguity of confusing `{}` with `set`. Please add a comment after this line that it is a `dict`. See the suggestion in the below comment.

czgdp1807:
```suggestion
        obj.edge_weights = {} # dict
```

czgdp1807:
Are you participating through any open source program? If yes then please let us know so that appropriate label can be attached.

pydatastructs/graphs/adjacency_list.py
sidhu1012: It's good to have comments but python by default treat `{}` as dictionary, it doesn't seem confusing.

pydatastructs/miscellaneous_data_structures/queue.py
sidhu1012: But if `if` is calling `return` then removing `if-else` is better approach I guess.

czgdp1807: > better readability

pydatastructs/graphs/adjacency_list.py
tusharnankani: What is needed to be done as of now? 

> Please add a comment after this line that it is a `dict`.

This will be quite cumbersome.

> It's good to have comments but python by default treat `{}` as a dictionary, it doesn't seem confusing.

I agree^

czgdp1807: Adding one extra word won't do much harm but taking care of this in future may be problematic. Let's leave it as it is (without comments).

czgdp1807:
There are a bunch of `if-return` which need to be restored to `if-elif` groups.

pydatastructs/graphs/graph.py
```diff
@@ -62,7 +62,7 @@ def __new__(cls, *args, **kwargs):
             return obj
         else:
             raise NotImplementedError("%s implementation is not a part "
-                                      "of the library currently."%(implementation))
+                                        "of the library currently."%(implementation))
```
czgdp1807, line 65:
```suggestion
                                       "of the library currently."%(implementation))
```

PR #342, opened 2021-03-10
Added `make_root` in DSU 

Smit-create:
Added a method to modify the parent of set to which `key` belongs

```py
>>> dst = DisjointSetForest()
>>> dst.make_set(1)
>>> dst.make_set(2)
>>> dst.union(1, 2)
>>> dst.find_root(2).key
1
>>> dst.make_root(2)
>>> dst.find_root(2).key
2
```

### DIFF
--- a/pydatastructs/miscellaneous_data_structures/disjoint_set.py
+++ b/pydatastructs/miscellaneous_data_structures/disjoint_set.py
@@ -16,6 +16,9 @@ class DisjointSetForest(object):
     >>> dst.union(1, 2)
     >>> dst.find_root(2).key
     1
+    >>> dst.make_root(2)
+    >>> dst.find_root(2).key
+    2
 
     References
     ==========
@@ -74,3 +77,18 @@ class DisjointSetForest(object):
 
             y_root.parent = x_root
             x_root.size += y_root.size
+
+    def make_root(self, key):
+        """
+        Finds the set to which the key belongs
+        and makes it as the root of the set.
+        """
+        if self.tree.get(key, None) is None:
+            raise KeyError("Invalid key, %s"%(key))
+
+        current_root = self.find_root(key)
+        if current_root.key != key:
+            key_set = self.tree[key]
+            current_root.parent = key_set
+            key_set.size = current_root.size
+            key_set.parent = key_set
--- a/pydatastructs/miscellaneous_data_structures/disjoint_set.py
+++ b/pydatastructs/miscellaneous_data_structures/disjoint_set.py
@@ -86,9 +86,20 @@ class DisjointSetForest(object):
         if self.tree.get(key, None) is None:
             raise KeyError("Invalid key, %s"%(key))
 
-        current_root = self.find_root(key)
-        if current_root.key != key:
-            key_set = self.tree[key]
-            current_root.parent = key_set
-            key_set.size = current_root.size
+        key_set = self.tree[key]
+        if key_set.parent is not key_set:
+            current_parent = key_set.parent
+            # Remove this key subtree size from all its ancestors
+            while current_parent.parent is not current_parent:
+                current_parent.size -= key_set.size
+                current_parent = current_parent.parent
+
+            all_set_size = current_parent.size # This is the root node
+            current_parent.size -= key_set.size
+
+            # Make parent of current root as key
+            current_parent.parent = key_set
+            # size of new root will be same as previous root's size
+            key_set.size = all_set_size
+            # Make parent of key as itself
             key_set.parent = key_set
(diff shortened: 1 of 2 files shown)


Smit-create:
@czgdp1807 I was thinking if we could change the implementation of the current DSU from maps to arrays(take `n`(size of array) as an argument while creating the instance) as it is costing an additional factor O(logN) in time complexity

czgdp1807:
In Python `dict` is a hashing based data structure It's counterpart in C++ is unordered_map. `map` in C++ is based on red-black tree and is not hashing based hence, it costs O(lg N) overhead.

pydatastructs/miscellaneous_data_structures/tests/test_disjoint_set.py
```diff
@@ -15,9 +15,15 @@ def test_DisjointSetForest():
...
+    assert dst.find_root(8).key == 8
     assert dst.find_root(3) == dst.find_root(4)
     assert dst.find_root(7).key == 7
-
+    dst.make_root(2)
     assert raises(KeyError, lambda: dst.find_root(9))
+    assert raises(KeyError, lambda: dst.make_root(19))
     dst.union(3, 1)
-    assert dst.find_root(3).key == 1
+    assert dst.find_root(3).key == 2
```
czgdp1807, line 29: Please add a new test in the same function. From here it appears like some existing tests are modified.

pydatastructs/miscellaneous_data_structures/disjoint_set.py
```diff
@@ -74,3 +77,18 @@ def union(self, key1, key2):
 
             y_root.parent = x_root
             x_root.size += y_root.size
+
+    def make_root(self, key):
```
czgdp1807, line 81: The feature looks good. Could you share an example (book/lecture notes examples will work) where this is used/required?

Smit-create: Presently, I don't remember the exact the example but, we can consider this example https://cp-algorithms.com/data_structures/disjoint_set_union.html#toc-tgt-10 wherein we generally make root which Is largest element in set

czgdp1807: I see. That's great. Could you solve one such problem using DSU API in `pydatastructs` and show the code here?

Smit-create:
@czgdp1807 Looks good to go?

pydatastructs/miscellaneous_data_structures/disjoint_set.py
Smit-create: I will trying finding the exact problem I solved using this method and show that

Smit-create: Well, this is the problem where I used the above method: https://codeforces.com/contest/1133/problem/F2, and the submission of mine can be found at: https://codeforces.com/contest/1133/submission/109674863

Smit-create:
@czgdp1807 Looks good to go?

pydatastructs/miscellaneous_data_structures/disjoint_set.py
```diff
@@ -74,3 +77,18 @@ def union(self, key1, key2):
...
+        and makes it as the root of the set.
+        """
+        if self.tree.get(key, None) is None:
+            raise KeyError("Invalid key, %s"%(key))
+
+        current_root = self.find_root(key)
+        if current_root.key != key:
+            key_set = self.tree[key]
+            current_root.parent = key_set
+            key_set.size = current_root.size
```
czgdp1807, line 93: I think the size of `current_root` should change. Just take the following example,
```python
>>> from pydatastructs import *
>>> dst = DisjointSetForest()
>>> dst.make_set(1)
>>> dst.make_set(2)
>>> dst.make_set(3)
>>> dst.tree[1].size
1
>>> dst.tree[2].size
1
>>> dst.tree[3].size
1
>>> dst.union(2, 3)
>>> dst.tree[1].size
1
>>> dst.tree[2].size
2
>>> dst.tree[3].size
1
>>> dst.make_root(3)
>>> dst.tree[1].size
1
>>> dst.tree[2].size # should have been 1 because now 2 is pointing to 3
2
>>> dst.tree[3].size
2
```
I think this inconsistency can cause problems when root is shifted a lot of times to different elements of set. We should keep things correct wherever possible.

Smit-create: > I think this inconsistency can cause problems when root is shifted a lot of times to different elements of set. We should keep things correct wherever possible.

This can be recitifed using re-rooting technique: https://wiki.algo.is/Rerooting%20technique. Actually while changing the root I didn't tried to maintain tree/subtree structure and just focused on the size of total elements belonging to that set. However using above technique we can re-root by maintaing tree/subtree structure too

czgdp1807: I think we should have it. We may need to track the nodes whose size will change due to change of root. It may or not be that tricky. Try to do it by just using the parent pointers.

czgdp1807:
LGTM. Merging. Thanks.

PR #344, opened 2021-03-11
Added a function to check ordering in ODA and DODA

Nora2412:
Appended algorithms file with is_sorted function. This will be useful to
check whether the array given is sorted or not.

### DIFF
--- a/pydatastructs/linear_data_structures/algorithms.py
+++ b/pydatastructs/linear_data_structures/algorithms.py
@@ -818,7 +818,6 @@ def is_ordered(array, start, end, comp):
     """
     if comp is None:
         comp = lambda a, b: a < b
-        
     for i in range(start + 1, end + 1):
         if(comp(array[i], array[i - 1])):
             return False
--- a/pydatastructs/linear_data_structures/algorithms.py
+++ b/pydatastructs/linear_data_structures/algorithms.py
@@ -798,20 +798,19 @@ def is_ordered(array, **kwargs):
 
     array: Array
         The array which is to be checked for having
-        specified ordered among its elements.
+        specified ordering among its elements.
     start: int
-        The starting index of the portion
-        which is to be sorted.
+        The starting index of the portion of the array
+        under consideration.
         Optional, by default 0
     end: int
-        The ending index of the portion which
-        is to be sorted.
+        The ending index of the portion of the array
+        under consideration.
         Optional, by default the index
         of the last position filled.
     comp: lambda/function
         The comparator which is to be used
-        for sorting. If the function returns
-        False then only swapping is performed.
+        for specifying the desired ordering.
         Optional, by default, less than or
         equal to is used for comparing two
         values.
--- a/pydatastructs/linear_data_structures/algorithms.py
+++ b/pydatastructs/linear_data_structures/algorithms.py
@@ -15,7 +15,7 @@ __all__ = [
     'cocktail_shaker_sort',
     'quick_sort',
     'longest_common_subsequence',
-    'is_sorted'
+    'is_ordered'
 ]
 
 def _merge(array, sl, el, sr, er, end, comp):
@@ -789,9 +789,9 @@ def longest_common_subsequence(seq1: OneDimensionalArray, seq2: OneDimensionalAr
 
     return OneDimensionalArray(seq1._dtype, check_mat[row][col][-1])
 
-def is_sorted(array, start, end):
+def is_ordered(array, start, end, comp):
     """
-    Checks whether the given array is sorted or not.
+    Checks whether the given array is ordered or not.
 
     Parameters
     =========
@@ -807,16 +807,19 @@ def is_sorted(array, start, end):
     Examples
     ========
 
-    >>> from pydatastructs import OneDimensionalArray, is_sorted
+    >>> from pydatastructs import OneDimensionalArray, is_ordered
     >>> arr = OneDimensionalArray(int,[4,3,2,1])
-    >>> is_sorted(arr,0,3)
+    >>> is_ordered(arr,0,3, None)
     False
     >>> arr1 = OneDimensionalArray(int,[1,2,3])
-    >>> is_sorted(arr1,0,2)
+    >>> is_ordered(arr1,0,2, None)
     True
 
     """
-    for i in range(start+1,end+1):
-        if(array[i]<array[i-1]):
+    if comp is None:
+        comp = lambda a, b: a < b
+
+    for i in range(start + 1, end + 1):
+        if(comp(array[i], array[i - 1])):
             return False
     return True
--- a/pydatastructs/linear_data_structures/algorithms.py
+++ b/pydatastructs/linear_data_structures/algorithms.py
@@ -808,29 +808,17 @@ def is_ordered(array, start, end, comp):
     ========
 
     >>> from pydatastructs import OneDimensionalArray, is_ordered
-<<<<<<< HEAD
-    >>> arr = OneDimensionalArray(int,[4,3,2,1])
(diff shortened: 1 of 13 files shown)


pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -787,3 +788,35 @@ def longest_common_subsequence(seq1: OneDimensionalArray, seq2: OneDimensionalAr
...
+
+    end : int
+        The ending index of the portion of array to be checked.
+
+    Examples
+    ========
+
+    >>> from pydatastructs import OneDimensionalArray, is_sorted
+    >>> arr = OneDimensionalArray(int,[4,3,2,1])
+    >>> is_sorted(arr,0,3)
```
Smit-create, line 812:
```suggestion
    >>> arr = OneDimensionalArray(int, [4, 3, 2, 1])
    >>> is_sorted(arr, 0, 3)
```
Make suggested changes wherever appicable. Add a space after comma

pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -787,3 +788,35 @@ def longest_common_subsequence(seq1: OneDimensionalArray, seq2: OneDimensionalAr
...
+    False
+    >>> arr1 = OneDimensionalArray(int,[1,2,3])
+    >>> is_sorted(arr1,0,2)
+    True
+
+    """
+    for i in range(start+1,end+1):
+        if(array[i]<array[i-1]):
```
Smit-create, line 820:
```suggestion
    for i in range(start + 1, end + 1):
        if(array[i] < array[i - 1]):
```

Smit-create (commented):
Please apply this changes wherever required

czgdp1807:
This is a bit trivial. I mean checking if an array sorted or not in O(N) time isn't really helpful. For keeping time complexity constant we may need to maintain a state inside ODA class to track when the sort function was called on an array object. More or less it would be just like maintaining the history of the array. We can have this feature as something extra.

pydatastructs/linear_data_structures/algorithms.py
czgdp1807, line 820: This will not work with `DODA` (the dynamic arrays). An operator is defined in `_comp` (or something similar) in `misc_util.py`. Please use that here. 
Moreover, we should allow a custom comparator (defaulting to `<=`) to check if an array is ordered according to some criteria specified by the passed comparator. The function can also be renamed to `is_ordered`.

Smit-create:
> For keeping time complexity constant we may need to maintain a state inside ODA class to track when the sort function was called on an array object

What if the user has'nt call sort function and has already given an sorted array as input?

> This is a bit trivial. I mean checking if an array sorted or not in O(N) time isn't really helpful.

Presently, C++, also uses linear time complexity to check sortedness: http://www.cplusplus.com/reference/algorithm/is_sorted/

> An operator is defined in `_comp` (or something similar) in `misc_util.py`. Please use that here.

This is something that would be good to have instead of using operators directly

Smit-create:
Taking references from https://en.cppreference.com/w/cpp/algorithm/lower_bound and https://en.cppreference.com/w/cpp/algorithm/is_sorted_until:
We could do something like this:
```py
def is_sorted(array, start, end, cmp):
    # cmp is a function that takes (a, b) and returns true if a is strictly less than b (a<b)
    if cmp is None:
         cmp = lambda a, b: (a < b) # we will use python defined <
    # algo
    return True/False
```
Similar thing can be done on #351 
What do you think @czgdp1807 ?

pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -787,3 +788,35 @@ def longest_common_subsequence(seq1: OneDimensionalArray, seq2: OneDimensionalAr
                     check_mat[i][j] = check_mat[i][j-1]
 
     return OneDimensionalArray(seq1._dtype, check_mat[row][col][-1])
+
+def is_sorted(array, start, end):
```
czgdp1807, line 792: Please rename this function to `is_ordered` and accept an extra parameter `comp`.

czgdp1807:
> What do you think @czgdp1807 ?

We should add `is_ordered` as it is more generic and allows more flexibility.

pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -787,3 +788,35 @@ def longest_common_subsequence(seq1: OneDimensionalArray, seq2: OneDimensionalAr
...
+    >>> is_ordered(arr,0,3)
+    False
+    >>> arr1 = OneDimensionalArray(int,[1,2,3])
+    >>> is_ordered(arr1,0,2)
+    True
+
+    """
+    for i in range(start + 1, end + 1):
+        if(array[i] < array[i - 1]):
+            return False
```
Smit-create, line 821:
```suggestion
    """
    if comp is None:
        comp = lambda a, b: a < b
    for i in range(start + 1, end + 1):
        if(comp(array[i], array[i - 1])):
            return False
```

pydatastructs/linear_data_structures/tests/test_algorithms.py
```diff
@@ -126,3 +126,36 @@ def test_longest_common_sequence():
...
+
+    arr3 = ODA(int, [0, -1, -2, -3, -4, 4], None)
+    output = is_ordered(arr3, 0, 5)
+    assert output == expected_result
+
+    expected_result = True
+
+    arr4 = ODA(int, [1, 2, 3, 4, 5, 6, 7, 8, 9, 10], None)
+    output = is_ordered(arr4, 0, 9)
+    assert output == expected_result
```
Smit-create, line 161: Also add test cases for custom defined `comp`

Smit-create:
@Nora2412  Please use `python -m pytest` to check locally which test is failing (presently due to trailing space/tab on some line)

Smit-create:
If you have trailing whitespace it will show errors. This one will fix unwanted spaces:
$ ./bin/strip_whitespace <file>

pydatastructs/linear_data_structures/tests/test_algorithms.py
```diff
@@ -126,3 +126,36 @@ def test_longest_common_sequence():
...
     assert str(output) == '[]'
+
+def test_is_sorted():
+    ODA = OneDimensionalArray
+
+    expected_result = True
+
+    arr = ODA(int,[1,2,5,6])
+    output = is_sorted(arr,0,len(arr)-1)
+    assert output == expected_result
```
Smit-create, line 137:
```suggestion
    arr = ODA(int, [1, 2, 5, 6])
    output = is_sorted(arr, 0, len(arr) - 1)
    assert output == expected_result
```
Apply this change whereever necessary

pydatastructs/linear_data_structures/tests/test_algorithms.py
```diff
@@ -126,3 +126,36 @@ def test_longest_common_sequence():
...
+
+    arr3 = ODA(int,[0,-1,-2,-3,-4,4])
+    output = is_sorted(arr3,0,5)
+    assert output == expected_result
+
+    expected_result = True
+
+    arr4 = ODA(int,[1,2,3,4,5,6,7,8,9,10])
+    output = is_sorted(arr4,0,9)
+    assert output == expected_result
```
Smit-create, line 161: Also add tests for user-defined `comp`

Smit-create (commented):
This looks good. Please address few suggestions

Nora2412:
@Smit-create can you please review and let me know if any further changes are required.

czgdp1807:
Will merge once the tests pass.

PR #347, opened 2021-03-11
Implemented dijkstras' Algorithm

avishkar2001:
Implementation of Graph!

Given a graph and a source vertex in the graph, find the shortest paths from source to all vertices in the given graph-Dijkstra's shortest path algorithm!!

The code finds the shortest distances from the source to all vertices. If we are interested only in the shortest distance from the source to a single target, we can break them for a loop when the picked minimum distance vertex is equal to the target 

If we are interested only in the shortest distance from the source to a single target, we can break them for a loop when the picked minimum distance vertex is equal to the target!

### DIFF
--- a/pydatastructs/graphs/dijkstras_algorithm.py
+++ b/pydatastructs/graphs/dijkstras_algorithm.py
@@ -1,0 +1,36 @@
+import matplotlib.pyplot as plt
+import networkx as nx
+
+
+g = nx.Graph()
+
+
+for i in range(1, 7):
+    g.add_node(i)
+
+print(g.nodes())
+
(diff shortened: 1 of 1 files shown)


Smit-create:
This is already present: 
https://github.com/codezonediitj/pydatastructs/blob/a21d00ebc909db7cd9439fe4f31fde244f327404/pydatastructs/graphs/algorithms.py#L737

Moreover, please read the guidelines at: https://github.com/codezonediitj/pydatastructs/wiki/Plan-of-Action-for-Adding-New-Data-Structures. Also, it will be good if you open an issue and dicuss with us your ideas and implementation before raising a PR

PR #388, opened 2021-05-11
Sparse table class and tests added

sHiVaNgI821:
Added a class for a sparse table, that can be used to answer range queries on a given array efficiently. It requires a precomputation that takes O(nlogn) time and space. It is added to the miscellaneous_data_structures directory.

### DIFF
--- a/pydatastructs/miscellaneous_data_structures/sparse_table.py
+++ b/pydatastructs/miscellaneous_data_structures/sparse_table.py
@@ -1,0 +1,82 @@
+import math
+
+__all__ = ['sparseTable']
+
+
+class sparseTable:
+    """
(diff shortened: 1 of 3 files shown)


Smit-create:
Also, please link the issues that are all related to sparse table so that we can have a track of them. Basic structure and API looks fine to me.

pydatastructs/miscellaneous_data_structures/sparse_table.py
```diff
@@ -0,0 +1,82 @@
+import math
+
```
Smit-create, line 2: Can be removed?

sHiVaNgI821: Yes, sorry forgot to remove it

PR #391, opened 2021-05-17
Lowest common ancestor added

sHiVaNgI821:
Added the lowest_common_ancestor method that returns the lowest common ancestor of two given vertices in a graph (an ancestor that is common to both vertices and farthest from the root vertex).

### DIFF
--- a/pydatastructs/graphs/algorithms.py
+++ b/pydatastructs/graphs/algorithms.py
@@ -11,7 +11,7 @@ from pydatastructs.miscellaneous_data_structures import (
 from pydatastructs.graphs.graph import Graph
 from pydatastructs.linear_data_structures.algorithms import merge_sort_parallel
 from pydatastructs import PriorityQueue
-
+import math
 __all__ = [
     'breadth_first_search',
     'breadth_first_search_parallel',
@@ -22,7 +22,8 @@ __all__ = [
     'shortest_paths',
     'all_pair_shortest_paths',
     'topological_sort',
-    'topological_sort_parallel'
+    'topological_sort_parallel',
+    'lowest_common_ancestor'
 ]
 
 Stack = Queue = deque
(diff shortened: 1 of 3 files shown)


czgdp1807:
@sHiVaNgI821 Please try to cover the read lines as well in https://app.codecov.io/gh/codezonediitj/pydatastructs/compare/391/diff
Are they checks for corner cases? If so, then please add tests covering them.

pydatastructs/graphs/algorithms.py
```diff
@@ -993,3 +995,110 @@ def _job(graph: Graph, u: str):
...
+    >>> graph.add_edge('v_3', 'v_5')
+    >>> lowest_common_ancestor(graph, 'v_2', 'v_5', 'binary_lifting')
+    'v_1'
+    >>> lowest_common_ancestor(graph, 'v_4', 'v_5', 'binary_lifting')
+    'v_3'
+
+    References
+    ==========
+
+    .. [1] https://en.wikipedia.org/wiki/Topological_sorting#Kahn's_algorithm
```
czgdp1807, line 1047: Are binary lifting and Kahn's algorithm the same thing?

pydatastructs/graphs/algorithms.py
```diff
@@ -993,3 +995,110 @@ def _job(graph: Graph, u: str):
...
+    """
+    import pydatastructs.graphs.algorithms as algorithms
+    func = "_" + algorithm + "_" + graph._impl
+    if not hasattr(algorithms, func):
+        raise NotImplementedError(
+            "Currently %s algorithm isn't implemented for "
+            "performing topological sort on %s graphs." % (algorithm, graph._impl))
```
czgdp1807, line 1054: I think you intend to refer to lowest common ancestor.

pydatastructs/graphs/algorithms.py
```diff
@@ -11,6 +11,7 @@
 from pydatastructs.graphs.graph import Graph
 from pydatastructs.linear_data_structures.algorithms import merge_sort_parallel
 from pydatastructs import PriorityQueue
+import math
```
czgdp1807, line 14: Is importing math necessary for binary lifting?

sHiVaNgI821: Yes I suppose. I have used the log2 function from the math library in my code.

pydatastructs/graphs/algorithms.py
sHiVaNgI821: No sorry I had to change the reference used. I'll do it right away.

PR #443, opened 2021-12-01
Migrate the project to Poetry

HarshCasper:
## Description

This PR:

- Migrates the project from setuptools to Poetry
- Make apt changes for the project to build and test on GHA

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,22 +28,18 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Upgrade pip version
-        run: |
-          python -m pip install --upgrade pip
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
 
-      - name: Install requirements
-        run: |
-          pip install -r requirements.txt
-          pip install -r docs/requirements.txt
+      - name: 'Install dependencies'
+        run: poetry install
 
       - name: Build package
-        run: |
-          python -m pip install .
+        run: poetry build
 
       - name: Run tests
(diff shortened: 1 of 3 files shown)


czgdp1807:
Is `poetry.lock` file necessary. It has a lot of lines.

HarshCasper:
Hi @czgdp1807 

This is my first attempt at migrating the project to Poetry. Please have a look and let me know the apt changes required. Based on my guesswork, these are the things that would be left:

- Documenting new changes
- Publishing the package on poetry
- Removing `setup.py` and `requirements.txt`

HarshCasper:
> Is `poetry.lock` file necessary. It has a lot of lines.

It is required to lock all the developers to the same set of dependencies as everyone else is using. More about this on the Docs here: https://python-poetry.org/docs/basic-usage/#installing-with-poetrylock

czgdp1807:
Can we reduce the dependencies there then? We only use the following. Python should be a minimum of 3.8 from 1.0.1.

```
sphinx==4.2.0
sphinx-readable-theme==1.3.0
myst_nb==0.13.1
codecov
pytest-cov
```

czgdp1807:
> * Removing `setup.py` and `requirements.txt`

Let's keep `setuptools` there. Some new contributors might not know about poetry so having `setuptools` will help them. Having `poetry` is also good for those who want to try something new or not really that into `setuptools`.

czgdp1807:
I would suggest having two workflows, one using `setuptools` and the other using `poetry`. Meanwhile this should also be fixed, #453

Issue #446, opened 2021-12-04
To incorporate Searching Algorithms for Linear Data Structures


EmperorArthurIX:
#### Description of the problem
The codebase is very extensive and well-designed; however, it does not seem to include searching algorithms for linear data structures as far as I could observe. I believe addition of searching algorithms would be a good improvement to the project. Certain algorithms I have thought can be included are:
> - Linear Search(array, number)
>
> - Binary Search(array, number, unsorted_input=False)
>
> - Jump Search(array, number, unsorted_input=False)

The above mentioned declarations are suggested, where the parameter `unsorted_input` can be used to explicitly learn from the user whether or not the input `array` is sorted beforehand. If `True`, we must first sort the array within the search function, then search. 

#### Comments

With certain modifications, we **may** even allow to return the sorted array along with search result, but I have no idea on how that will be done at the moment.

PS: I am currently working on the Radix Sort issue that is already open, and after that reaches a satisfactory stage, I wish to go on with this one. Discussions and suggestions are welcome.

HarsheetKakar:
With binary search and jump search in mind we can also have a `SortedOneDimentionalArray` as data structure which would take `OneDimentionalArray` as input and would dynamically choose the sorting algorithm based on how far the array is already sorted which would come from its [Hamming Distance](http://www.cse.psu.edu/~sxr48/pubs/TestingSortednessEncyclopedia.pdf) (wheather it would be `insertionSort` which works best for nearly sorted arrays and `quickSort` if the array is not sorted).

czgdp1807:
Interesting idea. I think the idea of `SortedOneDimensionalArray` should be discusses in a separate issue. Is there any prior work (like Wikipedia article, papers, or lecture notes available online) which can provide clarity and examples?

HarsheetKakar:
I have moved the conversation to Gitter once we reach a consensus I will create a separate issue.

czgdp1807:
> Gitter

It's inactive. Folks use Discord these days a lot. Github issue would be the best place to discuss as it is accessible by anyone in the open source community.

czgdp1807:
And may be we should provide a function that should take an array as input and suggest one of the pydatastructs APIs as output. The time complexity should be O(n) because anything more won't be worth it. The constant in O(n) should also be very small (we can benchmark for that).

czgdp1807:
[...]

PR #454, opened 2021-12-08
Included linear + binary search for 1D static arrays

skullVoid:

#446 

A search function is added. 
It has 2 necessary parameters(1. element to search, 2. order of the array) and one optional (3. comparator)
1. what to search
2. positive values (int or float allowed, anything else raises type error exception) mean ascending order, negative is descending, 0 is unsorted
3. function as a parameter; this is for when the usual >, < do not work on the datatype of the array

### DIFF
--- a/pydatastructs/linear_data_structures/algorithms.py
+++ b/pydatastructs/linear_data_structures/algorithms.py
@@ -1409,8 +1409,7 @@ def linear_search(array, elem, **kwargs):
 
     >>> from pydatastructs import OneDimensionalArray, linear_search
     >>> arr = OneDimensionalArray(int,[3, 2, 1])
-    >>> out = linear_search(arr, 2)
-    >>> print(out)
+    >>> linear_search(arr, 2)
     1
 
     References
@@ -1470,8 +1469,7 @@ def binary_search(array, elem, **kwargs):
 
     >>> from pydatastructs import OneDimensionalArray, binary_search
     >>> arr = OneDimensionalArray(int,[1, 2, 3, 5, 10, 12])
-    >>> out = binary_search(arr, 5)
-    >>> print(out)
+    >>> binary_search(arr, 5)
     3
 
     References
@@ -1487,7 +1485,7 @@ def binary_search(array, elem, **kwargs):
     Left = start
     Right = end
     while Left <= Right:
-        Middle = int((Left+Right)/2)
+        Middle = Left//2 + Right//2 + Left % 2 * Right % 2
         if array[Middle] == elem:
                 return Middle
         if comp(array[Middle], elem):
@@ -1539,8 +1537,7 @@ def jump_search(array, elem, **kwargs):
 
     >>> from pydatastructs import OneDimensionalArray, jump_search
     >>> arr = OneDimensionalArray(int,[1, 2, 3, 5, 10, 12])
-    >>> out = linear_search(arr, 5)
-    >>> print(out)
+    >>> linear_search(arr, 5)
     3
 
     References
--- a/pydatastructs/linear_data_structures/algorithms.py
+++ b/pydatastructs/linear_data_structures/algorithms.py
@@ -1374,18 +1374,113 @@ def bubble_sort(array, **kwargs):
     return array
 
 def linear_search(array, elem, **kwargs):
+    """
+    Implements linear search algorithm.
+
+    Parameters
+    ==========
+
+    array: OneDimensionalArray
+        The array which is to be sorted.
+    elem: same as that of the array
+    start: int
+        The starting index of the portion
+        which is to be searched.
+        Optional, by default 0
+    end: int
+        The ending index of the portion which
+        is to be searched.
+        Optional, by default the index
+        of the last position filled.
+    backend: pydatastructs.Backend
+        The backend to be used.
+        Optional, by default, the best available
+        backend is used.
+
+    Returns
+    =======
+
+    output: int
+        The index of elem if found.
+        If not found, returns -1.
+
+    Examples
+    ========
+
+    >>> from pydatastructs import OneDimensionalArray, linear_search
+    >>> arr = OneDimensionalArray(int,[3, 2, 1])
+    >>> out = linear_search(arr, 2)
+    >>> print(out)
+    1
+
+    References
+    ==========
+
+    .. [1] https://en.wikipedia.org/wiki/Linear_search
+    """
     raise_if_backend_is_not_python(
-            bubble_sort, kwargs.get('backend', Backend.PYTHON))
+            linear_search, kwargs.get('backend', Backend.PYTHON))
     start = kwargs.get('start', 0)
     end = kwargs.get('end', len(array) - 1)
     for i in range(start, end+1):
-        if array.data[i] == elem:
+        if array[i] == elem:
             return i
     return -1
 
 def binary_search(array, elem, **kwargs):
+    """
+    Implements binary search algorithm ON A SORTED ARRAY.
+
+    Parameters
+    ==========
+
+    array: OneDimensionalArray
+        The array which is to be sorted.
+    elem: same as that of the array
+    start: int
+        The starting index of the portion
+        which is to be searched.
+        Optional, by default 0
+    end: int
+        The ending index of the portion which
+        is to be searched.
+        Optional, by default the index
+        of the last position filled.
+    comp: lambda/function
+        The comparator which is to be used
+        for sorting. If the function returns
+        False then only swapping is performed.
+        Optional, by default, less than or
+        equal to is used for comparing two
+        values.
+    backend: pydatastructs.Backend
+        The backend to be used.
+        Optional, by default, the best available
+        backend is used.
+
+    Returns
+    =======
+
+    output: int
+        The index of elem if found.
+        If not found, returns -1.
+
+    Examples
+    ========
+
+    >>> from pydatastructs import OneDimensionalArray, binary_search
+    >>> arr = OneDimensionalArray(int,[1, 2, 3, 5, 10, 12])
+    >>> out = binary_search(arr, 5)
+    >>> print(out)
+    3
+
+    References
+    ==========
+
+    .. [1] https://en.wikipedia.org/wiki/Binary_search_algorithm
+    """
     raise_if_backend_is_not_python(
-            bubble_sort, kwargs.get('backend', Backend.PYTHON))
+            binary_search, kwargs.get('backend', Backend.PYTHON))
     start = kwargs.get('start', 0)
     end = kwargs.get('end', len(array) - 1)
     comp = kwargs.get("comp", lambda u, v: u <= v)
@@ -1393,51 +1488,81 @@ def binary_search(array, elem, **kwargs):
     Right = end
     while Left <= Right:
         Middle = int((Left+Right)/2)
-        if array.data[Middle] is elem:
+        if array[Middle] == elem:
                 return Middle
-        if comp is None:
-            if array.data[Middle] < elem:
-                Left = Middle + 1
-            else:
-                Right = Middle - 1
+        if comp(array[Middle], elem):
+            Left = Middle + 1
         else:
-            if comp(array.data[Middle], elem):
-                Left = Middle + 1
-            else:
-                Right = Middle - 1
+            Right = Middle - 1
     return -1
 
 def jump_search(array, elem, **kwargs):
+    """
+    Implements jump search algorithm ON A SORTED ARRAY.
+
+    Parameters
+    ==========
+
+    array: OneDimensionalArray
+        The array which is to be sorted.
+    elem: same as that of the array
+    start: int
+        The starting index of the portion
+        which is to be searched.
+        Optional, by default 0
+    end: int
+        The ending index of the portion which
+        is to be searched.
+        Optional, by default the index
+        of the last position filled.
+    comp: lambda/function
+        The comparator which is to be used
+        for sorting. If the function returns
+        False then only swapping is performed.
+        Optional, by default, less than or
+        equal to is used for comparing two
+        values.
+    backend: pydatastructs.Backend
+        The backend to be used.
+        Optional, by default, the best available
+        backend is used.
+
+    Returns
+    =======
+
+    output: int
+        The index of elem if found.
+        If not found, returns -1.
+
+    Examples
+    ========
+
+    >>> from pydatastructs import OneDimensionalArray, jump_search
+    >>> arr = OneDimensionalArray(int,[1, 2, 3, 5, 10, 12])
+    >>> out = linear_search(arr, 5)
+    >>> print(out)
+    3
+
+    References
+    ==========
+
+    .. [1] https://en.wikipedia.org/wiki/Jump_search
+    """
     raise_if_backend_is_not_python(
             bubble_sort, kwargs.get('backend', Backend.PYTHON))
     start = kwargs.get('start', 0)
     end = kwargs.get('end', len(array) - 1)
-    comp = kwargs.get("comp", lambda u, v: u <= v)
-    step = int(sqrt(end - start))
+    comp = kwargs.get("comp", lambda u, v: u < v)
+    step = int(sqrt(end - start + 1))
     current_position = step
     prev = start
-    if comp is None:
-        while array.data[min(current_position, end)] < elem:
-            prev = current_position
-            current_position += step
-            if prev >= end:
-                return -1
-        while array.data[prev] < elem:
-            prev += 1
-            if prev == min(current_position, end):
-                return -1
-        if array.data[prev] == elem:
+    while comp(array[min(current_position, end)], elem):
+        prev = current_position
+        current_position += step
+        if prev > end:
+            return -1
+    while prev <= min(current_position, end):
+        if array[prev] == elem:
             return prev
-    else:
-        while comp(array.data[min(current_position, end)], elem):
-            prev = current_position
-            current_position += step
-            if prev > end:
-                return -1
-        while comp(array.data[prev], elem):
-            prev += 1
-            if prev == min(current_position, end):
-                return -1
-        if array.data[prev] == elem:
-            return prev
+        prev += 1
     return -1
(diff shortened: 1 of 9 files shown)


pydatastructs/linear_data_structures/arrays.py
```diff
@@ -126,7 +126,7 @@ def __new__(cls, dtype=NoneType, *args, **kwargs):
     @classmethod
     def methods(cls):
         return ['__new__', '__getitem__',
-        '__setitem__', 'fill', '__len__']
+        '__setitem__', 'fill', '__len__', 'search']
```
czgdp1807, line 129: Please make it a function (like place in `algorithms.py` in `linear_data_structures`) instead of a class method?

skullVoid: Pardon me, shall I include this directly? Since all this does is search in 1D arrays. I wanted to include this for dynamic arrays too, but that removal of items just makes a void location and does not shift the items in those void locations

skullVoid: Regardless, I will include a jump search as it was mentioned as an idea in the issue. Again, for 1D static arrays. Is that fine?

czgdp1807: That should be fine. I am planning to add `ImplicitArray` so it would be useful to have search functions for that as well. And anyways, conceptually, search is an algorithm that just reads the data from an array. It's not responsible for updating the state of `OneDimensionalArray` class, so having it as a method is not of much use.

skullVoid: Understood. So should I just move this change to algorithms.py along with addition of the jump search?

czgdp1807: Yes. Though let's focus only on one search algorithm for now. I think you are adding `binary_search`. Please make it that clear while naming the function. Like explicitly name the functions as `binary_search`, `jump_search`, `ternary_search`, etc.

skullVoid: Actually, it is a hybrid. The search depends on the sorted order parameter. Positive means ascending sorted, zero is unsorted and negative is descending sorted.

skullVoid: For 0 the search is simply linear, else it is binary. There is only a change of the comparison sign for descending array

czgdp1807: I don't think that's a good idea. The logic for each type of array should be encapsulated in a single function. The user can decide on their own what they might want to use. Please make separate functions for `binary_search`, `linear_search`, etc.

skullVoid: If you prefer that I should separate this hybrid function, then I shall do that

skullVoid: Ok understood

skullVoid: Uh sorry to ask this. Should I make a new PR after doing as told?

skullVoid: I pushed the new commit.

pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -1369,3 +1372,58 @@ def bubble_sort(array, **kwargs):
         array._modify(force=True)
 
     return array
+
+def linear_search(array, elem):
+    for i in range(array._size):
```
czgdp1807, line 1377: Is `array.size` not working? We shouldn't access private attributes (the ones having `_` as prefix) unless really needed.

pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -1369,3 +1372,58 @@ def bubble_sort(array, **kwargs):
...
+        if array._data[i] == elem:
+            return i
+    return -1
+
+def binary_search(array, elem, comp = None):
+    L = 0
+    R = array._size - 1
```
czgdp1807, line 1384: `left`, `right`?

pydatastructs/linear_data_structures/arrays.py
```diff
@@ -148,6 +148,7 @@ def fill(self, elem):
 
     def __len__(self):
         return self._size
+
```
czgdp1807, line 151:
```suggestion
```

czgdp1807 (dismissed):

pydatastructs/linear_data_structures/algorithms.py
skullVoid: no no. While seeing the code, I noticed that only _size is being used, so I used that too

pydatastructs/linear_data_structures/algorithms.py
skullVoid: Yes

pydatastructs/linear_data_structures/arrays.py
skullVoid: Sorry, I didnt get this

pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -1369,3 +1372,58 @@ def bubble_sort(array, **kwargs):
...
+            else:
+                R = M - 1
+        else:
+            if comp(array._data[M], elem):
+                L = M + 1
+            else:
+                R = M - 1
+    return -1
+
+def jump_search(array, elem, comp = None):
```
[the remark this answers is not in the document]
skullVoid: No, I am afraid of making mistakes since I dont have much expertise, so I did the way I was confident in.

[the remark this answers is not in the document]
skullVoid: Sorry for coming short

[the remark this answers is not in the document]
skullVoid: Are there any issues?

pydatastructs/linear_data_structures/algorithms.py
czgdp1807: Please make it explicit then. Always use full names instead of short hand notations. Not many key strokes will be saved while typing, `L` instead of `left`.

pydatastructs/linear_data_structures/arrays.py
czgdp1807: It's a suggestion to remove that extra line. It's not needed.

pydatastructs/linear_data_structures/algorithms.py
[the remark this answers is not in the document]
czgdp1807: Yes. All the APIs should be consistent to the maximum extent possible. Now, here, `start`, `end` and `backend` should be included and the implementation should be modified to use these parameters.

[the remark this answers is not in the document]
skullVoid: There is some git issue I cannot understand, I have created a new PR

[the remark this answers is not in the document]
czgdp1807: Well, you had to first pull the changes from `forArrays` branch. `git pull your_remote_name forArrays`. That will fetch the suggestion you applied here, then you will be able to push to your `forArrays` branch.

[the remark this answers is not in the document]
skullVoid: Is there still some issue? Since this is marked as a draft.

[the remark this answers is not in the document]
czgdp1807: Please add tests for all your APIs. See, `tests/test_algorithms.py` file under, `linear_data_structures`.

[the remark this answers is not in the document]
czgdp1807: Doc string are also to be added. See other algorithms in the same file.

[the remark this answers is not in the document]
skullVoid: On it

pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -1369,3 +1372,197 @@ def bubble_sort(array, **kwargs):
...
+        If not found, returns -1.
+
+    Examples
+    ========
+
+    >>> from pydatastructs import OneDimensionalArray, linear_search
+    >>> arr = OneDimensionalArray(int,[3, 2, 1])
+    >>> out = linear_search(arr, 2)
+    >>> print(out)
+    1
```
czgdp1807, line 1414:
```suggestion
    >>> linear_search(arr, 2)
    1
```
Above should work. Can you try?

pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -1369,3 +1372,197 @@ def bubble_sort(array, **kwargs):
...
+    """
+    raise_if_backend_is_not_python(
+            binary_search, kwargs.get('backend', Backend.PYTHON))
+    start = kwargs.get('start', 0)
+    end = kwargs.get('end', len(array) - 1)
+    comp = kwargs.get("comp", lambda u, v: u <= v)
+    Left = start
+    Right = end
+    while Left <= Right:
+        Middle = int((Left+Right)/2)
```
czgdp1807, line 1490: What about, `Left//2 + Right//2`?

pydatastructs/linear_data_structures/tests/test_algorithms.py
```diff
@@ -322,3 +322,24 @@ def test_next_prev_permutation():
...
+    array = ODA(int, [1, 2, 5, 7, 10, 29, 40])
+    for i in range(len(array)):
+        assert i == binary_search(array, array[i])
+
+def test_jump_search():
+    ODA = OneDimensionalArray
+
+    array = ODA(int, [1, 2, 5, 7, 10, 29, 40])
+    for i in range(len(array)):
+        assert i == jump_search(array, array[i])
```
czgdp1807, line 345: What about corner cases like, empty array, one element array? And see `_test_common_sort` above. Repetition of tests can be avoided by doing something similar here.

pydatastructs/linear_data_structures/algorithms.py
skullVoid: I think (Left + Right)//2 is fine but not too sure about the sum of floors, for example, when left = 3, right = 5(when both odd), it will give 3 as middle

pydatastructs/linear_data_structures/tests/test_algorithms.py
skullVoid: Are you asking for the test function of the actual algorithm written?

pydatastructs/linear_data_structures/algorithms.py
skullVoid: Yes Sure

skullVoid: I did not test it in conda environment so i am unfamiliar. But it works as intended, I checked in vs code.

pydatastructs/linear_data_structures/algorithms.py
czgdp1807: Well the loss won't be much. There are four cases,

1. Left - Odd, Right - Odd - Middle will be lesser (by 1) in Left//2 + Right//2 as compared to current approach.
2. Left - Odd, Right - Even or Right - Odd, Left - Even - Middle will be exactly the same as in current approach.
3. Left - Even, Right - Even - Middle will again be exactly the same as in current approach.

The reason is (Left + Right) might get too large for implicit arrays leading to slow down in performance (Python uses big integers for handling larger than 64 bit integers). So, basically if you can just add 1 for Left - Odd, Right - Odd in Left//2 + Right//2 and do only Left//2 + Right//2 for other cases then it should be fine. In fact,

`Middle = Left//2 + Right//2 + Left % 2 * Right % 2` should work. Worth trying.

skullVoid: Yes, this seems much better, I shall do this

pydatastructs/linear_data_structures/tests/test_algorithms.py
czgdp1807: Yes. You are using the same test for every search algorithm. So why not write a common test function, `_test_common_search(search_function_api)` and then in `test_jump_search` (for example),

```python
def test_jump_search():
    _test_common_search(jump_search)
```

czgdp1807:
See [this](https://app.codecov.io/gh/codezonediitj/pydatastructs/compare/454/diff). The read lines are not covered. Add tests for covering them as well.

pydatastructs/linear_data_structures/tests/test_algorithms.py
skullVoid: Right, I will do this

pydatastructs/linear_data_structures/algorithms.py
czgdp1807: I think only Python >= 3.5 is needed for testing. Doesn't matter whether it comes from conda, or VS Code.

czgdp1807: See the testing section of README.

skullVoid: Yes, I havent done it like that, I just ran sample codes to check. But it works

skullVoid:
> See [this](https://app.codecov.io/gh/codezonediitj/pydatastructs/compare/454/diff). The red lines are not covered. Add tests for covering them as well.

Its showing that bubble sort is not covered, but it is, I did not touch anything other than searching which I added

pydatastructs/linear_data_structures/algorithms.py
skullVoid: Ok, will do

czgdp1807:
> Its showing that bubble sort is not covered, but it is, I did not touch anything other than searching which I added

That's the red (lighter shade) of diff. If you scroll and click on `Click to load diff` then you will see dark red lines. These are the ones that you have to cover by adding tests.

skullVoid:
> > Its showing that bubble sort is not covered, but it is, I did not touch anything other than searching which I added
> 
> That's the red (lighter shade) of diff. If you scroll and click on `Click to load diff` then you will see dark red lines. These are the ones that you have to cover by adding tests.

Ok on it. But still, bubble sort and another line of importing is in deep red, what about it?

pydatastructs/linear_data_structures/tests/test_algorithms.py
```diff
@@ -323,23 +323,22 @@ def test_next_prev_permutation():
...
+        assert i == search(array, array[i])
+    
+    checker_array = [-1, -1, 2, 3, 4, 5, -1]
+    for i in range(len(array)):
+        assert checker_array[i] == search(array, array[i], start=2, end=5)
 
-def test_binary_search():
-    ODA = OneDimensionalArray
+def test_linear_search():
+    _test_common_search(test_linear_search)
```
czgdp1807, line 338:
```suggestion
    _test_common_search(linear_search)
```

pydatastructs/linear_data_structures/tests/test_algorithms.py
```diff
@@ -323,23 +323,22 @@ def test_next_prev_permutation():
...
 
-    array = ODA(int, [1, 2, 5, 7, 10, 29, 40])
-    for i in range(len(array)):
-        assert i == binary_search(array, array[i])
+def test_binary_search():
+    _test_common_search(test_binary_search)
```
czgdp1807, line 341:
```suggestion
    _test_common_search(binary_search)
```

pydatastructs/linear_data_structures/tests/test_algorithms.py
```diff
@@ -323,23 +323,22 @@ def test_next_prev_permutation():
...
 
 def test_jump_search():
-    ODA = OneDimensionalArray
-
-    array = ODA(int, [1, 2, 5, 7, 10, 29, 40])
-    for i in range(len(array)):
-        assert i == jump_search(array, array[i])
+    _test_common_search(test_jump_search)
```
czgdp1807, line 344:
```suggestion
    _test_common_search(jump_search)
```

pydatastructs/linear_data_structures/tests/test_algorithms.py
```diff
@@ -323,23 +323,22 @@ def test_next_prev_permutation():
         _, orig_array = next_permutation(prev_array)
         assert str(orig_array) == str(array)
 
-def test_linear_search():
+def _test_common_search(search):
```
czgdp1807, line 326:
```suggestion
def _test_common_search(search_func):
```

pydatastructs/linear_data_structures/tests/test_algorithms.py
```diff
@@ -323,23 +323,22 @@ def test_next_prev_permutation():
...
     ODA = OneDimensionalArray
 
     array = ODA(int, [1, 2, 5, 7, 10, 29, 40])
     for i in range(len(array)):
-        assert i == linear_search(array, array[i])
+        assert i == search(array, array[i])
+    
+    checker_array = [-1, -1, 2, 3, 4, 5, -1]
+    for i in range(len(array)):
+        assert checker_array[i] == search(array, array[i], start=2, end=5)
```
czgdp1807, line 335:
```suggestion
        assert checker_array[i] == search_func(array, array[i], start=2, end=5)
```

czgdp1807:
> bubble sort and another line of importing is in deep red, what about it?

Ignore. Though you can generate the same report locally if Codecov UI is confusing. See the testing section of README. Once you execute the command there, you will have to open the `index.html` file in `html/cov` in a browser.

skullVoid:
> > bubble sort and another line of importing is in deep red, what about it?
> 
> Ignore. Though you can generate the same report locally if Codecov UI is confusing. See the testing section of README. Once you execute the command there, you will have to open the `index.html` file in `html/cov` in a browser.

Can I use Codecov?

czgdp1807:
Sure.

skullVoid:
> Sure.

Is there any other issue?

czgdp1807:
I will push some changes may be on Sunday. I don't think there is anything else to be done from your side.

skullVoid:
> I will push some changes to your branch `forArrays` may be on Sunday. I don't think there is anything else to be done from your side.

Please let me know if there is any error. Shall I see if I can contribute something else in the project?

czgdp1807:
> Shall I see if I can contribute something else in the project?

Sure go ahead.

czgdp1807:
Thanks @skullVoid

skullVoid:
> Thanks @skullVoid

I didnt do much. Thank you for guiding me, this was a good experience for me as a first timer.

Issue #444, opened 2021-12-01
Versioning Tool/Script


czgdp1807:
#### Description of the problem

We need a tool/script that can update version number everywhere in the project by executing a single command. Following are the constraints,

1. We decide the version numbers manually.
2. We want to update to a new version number by executing a single command.
3. The way to do it should be simple - Say, `some-tool 1.0.1-dev 1.0.1-alpha`.

#### References/Other comments
cc: @HarshCasper

youknowqyh:
What does everywhere mean to be more specific? By command `grep 1.0.1-dev -r ./pydatastructs` I can see a lot of positions with `1.0.1-dev`. Do I need to change all these texts with `1.0.1-dev` to `1.0.1-alpha` for example?

czgdp1807:
The above is just an example. We will be releasing 6 months later (at the earliest), so we need to decide upon a tool that can update the version number everywhere in one go.

czgdp1807:
By everywhere I mean, where `1.0.1-dev` is present inside the project (README, code, documentation source) should be updated to say, `1.0.1-alpha` and so on.

youknowqyh:
Make sense. I'll work on it.

FamALouiz:
I would like to handle this task if it is still open.

PR #466, opened 2021-12-15
Implemented version tool

youknowqyh:
Implemented a version tool to update version number everywhere.
image

Implemented #444

### DIFF
--- a/version_tool.py
+++ b/version_tool.py
@@ -1,0 +1,33 @@
+import os
+import sys
+
+if len(sys.argv) != 3:
+    raise ValueError('original version number and new version number are both needed!')
+
+origin_version, new_version = sys.argv[1], sys.argv[2]
+
+# file_path contains all files'(README, code, documentation source) paths that may contain version number.
+file_path = ['README.md', 'setup.py']
+
+# Walk through directories to add all corresponding files' paths into file_path
+for root, _, files in os.walk('./pydatastructs'):
+    for file in files:
+
+        if file.endswith('.py'):
+            file_path.append(os.path.join(root, file))
+
+for root, dirs, files in os.walk('./docs/source'):
+    for file in files:
+        if file.endswith('.rst') or file.endswith('.py'):
+            file_path.append(os.path.join(root, file))
+
+# Update version number everywhere
+for path in file_path:
+    with open(path, 'r') as file:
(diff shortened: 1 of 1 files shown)


version_tool.py
```diff
@@ -0,0 +1,33 @@
...
+if len(sys.argv) != 3:
+    print('original version number and new version number are both needed!')
+else:
+    origin_version, new_version = sys.argv[1], sys.argv[2]
+
+    # file_path contains all files'(README, code, documentation source) paths that may contain version number.
+    file_path = ['README.md', 'setup.py']
+
+    # Walk through directories to add all corresponding files' paths into file_path
+    for root, _, files in os.walk('./pydatastructs'):
```
czgdp1807, line 13: May be we can re-use the following function to list all the files,

https://github.com/codezonediitj/pydatastructs/blob/97cb9fdd177f11b763c3f72ead11b4ba525ce4ed/pydatastructs/utils/tests/test_code_quality.py#L4

version_tool.py
```diff
@@ -0,0 +1,33 @@
...
+        for file in files:
+            if file.endswith('.rst') or file.endswith('.py'):
+                file_path.append(os.path.join(root, file))
+
+    # Update version number everywhere
+    for path in file_path:
+        with open(path, 'r') as file:
+            data = file.read()
+        if origin_version in data:
+            data = data.replace(origin_version, new_version)
```
czgdp1807, line 29: 👍🏻

czgdp1807:
Sorry for the delay in reviewing.

version_tool.py
youknowqyh: yeah, i see. But the root_path in `_list_files` doesn't make sense since it will go to '../../cur_file_path'.

So I try to modify the `_list_files` to pass one more argument, like `root_path=None`, so we can specify the root_path at will. 

Then one another issue occurred, you can see from the issue: ImportError #479

czgdp1807: > So I try to modify the `_list_files` to pass one more argument, like `root_path=None`, then we can specify the root_path at will.
> 
> Does adding such a modification in `_list_files` make sense?

Sounds good to me. `_list_files` is anyway a private function, so changing its API is not a problem.

Issue #376, opened 2021-04-22
Staircase Search


ShubhamGupta577:
#### Description of the problem
- Implement Staircase Search Algorithm.
- Used to search an element from a row-wise and column-wise sorted 2-D array.
- Time Complexity: O(max(n, m)), where n=number of rows, m= number of columns
- Space Complexity: O(1)
#### Example of the Problem
Dimensions of the array: 5 5
Element to be searched: 20
1 4 8 10 45
2 5 9 15 55
6 16 18 20 12
7 14 21 27 11
11 17 19 23 19

Output:
20 found at (2,3)

ShubhamGupta577:
@czgdp1807, Under GSSOC21 I would like to work on this issue. Please assign this issue to me.

kusumita29:
I would like to work on it under GSSOC'22. Please assign this issue to me.

czgdp1807:
@kusumita29 Please feel free to start working on this issue. We don't assign issues. Please read, https://github.com/codezonediitj/pydatastructs/wiki/Issue-Policy

kusumita29:
Thank you, Sir. I will start working on this issue.

mridul45:
As a GSSOC 23 contributor , Kindly allow me to contibute to this issue.

PR #498, opened 2022-03-14
Added StairCase Search for 2D Lists

Rajveer100:
Added StairCase Algo for 2D Lists in the linear data structures directory...


Rajveer100:
For Issue #376

czgdp1807:
How is segment tree related to stair case search?

czgdp1807:
How is segment tree related to stair case search?

PR #502, opened 2022-03-16
Add Useful Functions for 2D Lists

Rajveer100:
Added StairCase Search for 2D Lists under /linear_data_structures

### DIFF
--- a/pydatastructs/linear_data_structures/algorithms.py
+++ b/pydatastructs/linear_data_structures/algorithms.py
@@ -1757,3 +1757,31 @@ def stair_case_search(search_list, val):
                 return [i, j]
 
     return False
+
+def matrix_search(mat, val, strict=False):
+
+    if not any(mat):
+
+        return False
+
+    n, m = len(mat), len(mat[0])
+
+    if val < mat[0][0] or val > mat[n - 1][m - 1]:
+
+        return False
+
+    i, j = 0, m - 1
+
+    while i <= n - 1 and j >= 0:
+
+        if mat[i][j] == val:
+
+            return [i, j]
+        elif val < mat[i][j]:
+
+            j -= 1
+        else:
+
+            i += 1
+
+    return False
--- a/pydatastructs/linear_data_structures/algorithms.py
+++ b/pydatastructs/linear_data_structures/algorithms.py
@@ -1733,3 +1733,27 @@ def jump_search(array, value, **kwargs):
         prev += 1
 
     return None
+    
+def stair_case_search(search_list, val):
+    """
+    Implements Stair_Case Search Algorithm
+    Parameters
+    ==========
+    search_list: 2-D List.
+    val: Search Value.
+    Returns
+    =======
+    output: int
+        The index of the search element.
+    """
+    n, m = len(search_list), len(search_list)
+
+    for i in range(n):
+
+        for j in range(m):
+
+            if search_list[i][j] == val:
+
+                return [i, j]
+
+    return False


Rajveer100:
@czgdp1807 For issue number #376

pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -1733,3 +1733,27 @@ def jump_search(array, value, **kwargs):
...
+    """
+    n, m = len(search_list), len(search_list)
+
+    for i in range(n):
+
+        for j in range(m):
+
+            if search_list[i][j] == val:
+
+                return [i, j]
```
czgdp1807, line 1757: Well, this would take O(m*n) steps. This generalises to O(number_of_elements_in_the_input_container) for arbitrary dimensional array. `MultiDimensionalArray._data` in `arrays.py` represents data internally as a 1-D list. So, the above two nested loops would be equivalent to a single loop traversing `MultiDimensionalArray._data`. Hence, according to your implementation stair case search for 2-D arrays is same as linear search for 1-D arrays time complexity wise. I am unable to see the advantages of a special function like this.

Could you please explain how would it be beneficial to add a special function, `stair_case_search` for 2-D arrays?

czgdp1807: I see. Is there any algorithm for multi-dimensional arrays for this purpose (finding k-th smallest/largest element in a given dimension)? Wouldn't that be something unique and interesting?

Rajveer100: I feel like this topic wouldn't help much as a contribution, because the further restriction that is caused is that, array must be sorted...and that is pretty rare in most use cases...

Rajveer100: Maybe I should find some useful function for matrix related searching or any function that could actually help..

Rajveer100: What would you say?

Rajveer100: Is Matrix Chain Multiplication added to this repo? Maybe this could be little useful..

Rajveer100: > Let's think this way, if an array is sorted and then you have to answer a million queries after that. Answering each query in lesser time complexity would be preferred. So, in that context this contribution would be helpful. The problem is we need to figure something out for multi-dimensional arrays (something that works for more than 2-D). Can you share the algorithm that is defined for 2-D arrays?

Yes for query purpose, this is would help definitely, so yeah regarding the Algo, can I just directly commit it?
As it's actually very intuitive, it uses the fact that the 2-D array is sorted...

Rajveer100: Also regarding the Algo..There are two cases..
First case: When the matrix is strictly sorted where each row's element is smaller than it's previous row...
Second case: Each row is sorted but the strict condition may not be true..

So for the first case, we can binary search to achieve log(n)+log(m) time complexity...For the second case it would be linear..

czgdp1807: Can you give an example of both the cases?

Rajveer100: > Let's think this way, if an array is sorted and then you have to answer a million queries after that. Answering each query in lesser time complexity would be preferred. So, in that context this contribution would be helpful. The problem is we need to figure something out for multi-dimensional arrays (something that works for more than 2-D). Can you share the algorithm that is defined for 2-D arrays?

Also regarding the Algo..There are two cases..
First case: When the matrix is strictly sorted where each row's element is smaller than it's previous row...
Second case: Each row is sorted but the strict condition may not be true..

So for the first case, we can binary search to achieve log(n)+log(m) time complexity...For the second case it would be linear..

Rajveer100: Consider this example (for second case):

4 3
6 9 10
1 4 5
2 8 17
3 6 100

If you had to search 17...suppose..you would start top right...and keep going left of down depending on the value of search element..

Rajveer100: Added the commit for this case..you can have a look at the function...will also had for the strict case (strictly sorted)...soon.

czgdp1807: Do you know of any applications of stair case search?

Rajveer100: Well as you said, if we had several queries, we can answer them in O(Q*(N+M)) time in situations...and the strictly sorted version, would allow searching in O(Q*(log(N)+log(M))) time...It actually acts as a quick function which the user would have to manually do, and would save time...

czgdp1807: That's okay. But how will we use output of stair case search? I think it just produces the first index tuple where the given number/object is found? So, basically you can find out whether a given number/object exists in a matrix or not? What's the use of this information. If I have to answer million queries just to check whether a given number/object exists in a matrix or not. I would spend time to pre-process a matrix into a dict (key = matrix element, value = vector of indices) and then answer million queries using that dict. So, the complexity would be O(N*M + Q) instead of O(Q*(N + M)). If N = M = 1000, then O(N*M + Q) = (million * 2) but O(Q*(N + M)) = O(million * 2000). No? In fact, O(Q*(log(N) + log(M)) would be O(million * 2 * log(1000)). In fact, dict would help in providing all the indices where a given value would be present in a single query. Any thoughts?

Rajveer100: You are right...but the problem with python packages is that..it used a Hash Map..and that only guarantees an average case time of O(1), and can lead to bad hash collisions in worst cases..which would essentially make it linear...

And yes, use case related it's quite less..but that way many functions already existing have their specific situations...

czgdp1807: > it uses a Hash Map..and that only guarantees an average case time of O(1)

Not a big problem. You can define your own hash function (`__hash__` method). There are very good functions out there that one can use to lower collision. 

Overall, I am -1 for this stair case algorithm. Thanks for discussion anyways.

Rajveer100: I thought so...actually this wasn't my idea...as I mentioned, this was put because someone had put this issue..and I thought of solving that..because he/she might have put for a certain use case...

czgdp1807: No worries. I would still recommend to go for graph algorithms. That should be easy yet impactful work. Should require little discussion. Though, I would recommend to go through the exisiting implementations in `graph/algorithms.py`.

Issue #511, opened 2022-07-29
Porting `stack.py` to C++ backend 


czgdp1807:
#### Description of the problem

We need to port `stack.py` code in Python to C++ code. We did this for arrays before, 

1. Python code for arrays - https://github.com/codezonediitj/pydatastructs/blob/master/pydatastructs/linear_data_structures/arrays.py2. C++ code - https://github.com/codezonediitj/pydatastructs/tree/master/pydatastructs/linear_data_structures/_backend/cpp/arrays

You can follow a similar pattern for stacks. Feel free to ask any queries.

nubol23:
I want to work on this Issue

czgdp1807:
Sure. @nubol23. Feel free to go ahead and start working on it. :-)

PoulasthaMukherjee:
@czgdp1807 May I work on this as well?

patelalay231:
@czgdp1807 can I start working on it?

bobbysharma05:
@czgdp1807 Can I work on this issue for GSOC2025, It will also like my revision of Data Structures.

bobbysharma05:
@czgdp1807 So do we hav eto only make the Stack.py file or we also have to make all the stack files in c++. Just like it was done in the arrays

VW100:
@czgdp1807 I am a second-year student proficient in C++. Can I contribute to this project under GSoC 2025?

frHimanshu:
this issue is already solved, isnt it?

manitejagaddam:
Is the issue resolved i cant see any issues in the code base

0xChaitanya:
I think this issue is resolved, isn't it?

priti200:
Hi @czgdp1807

I was reading through this issue and saw that PR #591 was linked for adding the C++ backend support for stacks. I just wanted to check if everything from the original scope is now fully implemented and merged.

If there are still any gaps, improvements, or additional tests needed, I’d be happy to pick them up and help move this forward.

Please let me know how I can contribute 🙂

PR #512, opened 2022-08-02
Adding support for ``stack.ArrayStack`` in C++ backend

nubol23:
Adds cpp backend for the stack implementation.

Based on Issue #511

Work in progress

### DIFF
--- a/pydatastructs/linear_data_structures/_backend/cpp/arrays/DynamicOneDimensionalArray.hpp
+++ b/pydatastructs/linear_data_structures/_backend/cpp/arrays/DynamicOneDimensionalArray.hpp
@@ -8,7 +8,6 @@
 #include "DynamicArray.hpp"
 #include "OneDimensionalArray.hpp"
 #include "../../../../utils/_backend/cpp/utils.hpp"
-#include <iostream>
 
 typedef struct {
     PyObject_HEAD
--- a/pydatastructs/linear_data_structures/_backend/cpp/arrays/DynamicOneDimensionalArray.hpp
+++ b/pydatastructs/linear_data_structures/_backend/cpp/arrays/DynamicOneDimensionalArray.hpp
@@ -8,6 +8,7 @@
 #include "DynamicArray.hpp"
 #include "OneDimensionalArray.hpp"
 #include "../../../../utils/_backend/cpp/utils.hpp"
+#include <iostream>
 
 typedef struct {
     PyObject_HEAD
--- a/pydatastructs/miscellaneous_data_structures/_backend/cpp/stack/stack.cpp
+++ b/pydatastructs/miscellaneous_data_structures/_backend/cpp/stack/stack.cpp
@@ -3,27 +3,46 @@
 
 static struct PyModuleDef stack_struct = {
     PyModuleDef_HEAD_INIT,
-    .m_name = "_stack",
-    .m_doc = "Stack extension type",
-    .m_size = -1,
+    "_stack",
+    0,
+    -1,
+    NULL,
 };
 
 PyMODINIT_FUNC PyInit__stack(void) {
-    PyObject *stack;
+    Py_Initialize();
+    PyObject *stack = PyModule_Create(&stack_struct);
 
-    if (PyType_Ready(&ArrayStackType) < 0)
+    if (PyType_Ready(&ArrayStackType) < 0) {
         return NULL;
-
-    stack = PyModule_Create(&stack_struct);
-    if (stack == NULL)
-        return NULL;
-
+    }
     Py_INCREF(&ArrayStackType);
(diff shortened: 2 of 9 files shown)


pydatastructs/miscellaneous_data_structures/_backend/cpp/stack/stack.cpp
```diff
@@ -0,0 +1,48 @@
...
+        return NULL;
+    }
+    Py_INCREF(&DynamicArrayType);
+    PyModule_AddObject(stack, "DynamicArray", reinterpret_cast<PyObject*>(&DynamicArrayType));
+
+    if (PyType_Ready(&DynamicOneDimensionalArrayType) < 0) {
+        return NULL;
+    }
+    Py_INCREF(&DynamicOneDimensionalArrayType);
+    PyModule_AddObject(stack, "DynamicOneDimensionalArray", reinterpret_cast<PyObject*>(&DynamicOneDimensionalArrayType));
```
czgdp1807, line 44: The segmentation was being thrown because the array types weren't registered in the `stack` extension module. So there are two options,

1. Register them manually (as I did in the above). 
2. Import them at runtime via the APIs in https://docs.python.org/3/c-api/import.html.

For now approach in point 1 seems much easier and efficient because we pre-register everything that's going to be used in the module's implementation. So no need to import them during function calls. The second approach is messier as well (`PyImport_Import` every now and then whenever we want to use something from another C++ extension module).

czgdp1807:
In addition to https://github.com/codezonediitj/pydatastructs/pull/512#discussion_r939642998 I removed calls to `Py_INCREF`/`Py_XDECREF` as they sometimes give segmentation faults. Anyways, avoid the calls to these two unless extremely necessary, like if your code only works after adding them at specific places then feel free to add otherwise ignore.

pydatastructs/miscellaneous_data_structures/_backend/cpp/stack/stack.cpp
czgdp1807: The following code works for me for example with 4b028f158556f70ebc8bb144d0995b1faea70b4f commit.

```python
from pydatastructs.miscellaneous_data_structures._backend.cpp import _stack

array_stack = _stack.ArrayStack([1, 2, 3], int)
```

czgdp1807:
Great progress @nubol23. I am glad that you are able to figure things out on your own in Python C-API. Let me know if you get stuck somewhere, we will fix the problem. :-)

nubol23:
Thank you very much @czgdp1807 :) I added the remaining changes for the ArrayStack

pydatastructs/miscellaneous_data_structures/stack.py
```diff
@@ -53,18 +54,25 @@ class Stack(object):
...
-                kwargs.get('dtype', int))
+            items = kwargs.get('items', None)
+            dtype = kwargs.get('dtype', int)
+            if backend == Backend.CPP:
+                if items is None:
+                    return _stack.ArrayStack(dtype)
+                else:
+                    return _stack.ArrayStack(items, dtype)
+            else:
+                return ArrayStack(items, dtype)
```
czgdp1807, line 67: It would be great if we can handle this in C++ backend itself? Specifically I would just do `_stack.ArrayStack(**kwargs)` and then C++ backend should try to fetch `item`, `dtype` from `**kwargs` using `PyObject_GetItem` (confirm from Python C-API docs).

Can you try it out?

nubol23: Yep, we can send positional args and named args now

pydatastructs/linear_data_structures/_backend/cpp/arrays/DynamicOneDimensionalArray.hpp
```diff
@@ -8,6 +8,7 @@
 #include "DynamicArray.hpp"
 #include "OneDimensionalArray.hpp"
 #include "../../../../utils/_backend/cpp/utils.hpp"
+#include <iostream>
```
czgdp1807, line 11:
```suggestion
```

pydatastructs/miscellaneous_data_structures/stack.py
```diff
@@ -53,18 +54,22 @@ class Stack(object):
...
         if implementation == 'array':
-            return ArrayStack(
-                kwargs.get('items', None),
-                kwargs.get('dtype', int))
+            items = kwargs.get('items', None)
+            dtype = kwargs.get('dtype', int)
+            if backend == Backend.CPP:
+                return _stack.ArrayStack(items, dtype)
+            else:
+                return ArrayStack(items, dtype)
```
czgdp1807, line 64:
```suggestion

            return ArrayStack(items, dtype)
```

czgdp1807:
Great job @nubol23. 🎉🎉

PR #513, opened 2022-08-12
Adding support for graph.AdjacencyMatrix in C++ backend

nubol23:
Add graph cpp backend.

Currently working on AdjacencyMatrix implementation.


czgdp1807:
So, while adding support any data structure in C++ backend we need to make sure that there is no difference between the interface supported by C++ backend and Python backend. Internally both can follow different implementations. 

Since we have already implemented Graph in Python backend so we need to make sure that interface of every method defined in https://github.com/codezonediitj/pydatastructs/blob/master/pydatastructs/graphs/graph.py works with C++ backend as well. The pattern I notice is here that we pass nodes as input, call `str` upon these nodes and use the output to index the graph data structure.

So I would say let's follow a similar pattern for now in C++ backend as well, with a small tweak i.e., I would use `name` attribute instead of calling `str` on the input nodes which will return the name for these nodes. Then use these names to index the data structure.

Now, `AdjacencyMatrixCpp` can contain the following containers to store information related to nodes and edges,

1. `std::map<std::pair<std::string, std::string>, bool>` - Similarly to https://github.com/codezonediitj/pydatastructs/blob/8f419fd11ad73e9a60666d7834797cce321bb975/pydatastructs/graphs/adjacency_matrix.py#L26-L28.
2. `std::map<std::pair<std::string, std::string>, GraphEdgeCpp>` - Similar to https://github.com/codezonediitj/pydatastructs/blob/8f419fd11ad73e9a60666d7834797cce321bb975/pydatastructs/graphs/adjacency_matrix.py#L29

And in addition to `AdjancencyMatrixCpp` we will need following data structures in C++ backend,

1. `GraphNodeCpp` - Similar to https://github.com/codezonediitj/pydatastructs/blob/8f419fd11ad73e9a60666d7834797cce321bb975/pydatastructs/utils/misc_util.py#L364-L369.
2. `AdjancencyMatrixGraphNodeCpp` - Similar to https://github.com/codezonediitj/pydatastructs/blob/8f419fd11ad73e9a60666d7834797cce321bb975/pydatastructs/utils/misc_util.py#L434
3. `GraphEdgeCpp` - Similar to https://github.com/codezonediitj/pydatastructs/blob/8f419fd11ad73e9a60666d7834797cce321bb975/pydatastructs/utils/misc_util.py#L467

PR #514, opened 2022-10-07
Linear probing map

nubol23:
Adds Linear Probing Hash Map templated implementation in utils.

Currently support adition and search

TODO: support deletion


czgdp1807 (commented):
This is a nice addition. However I think we should expose this via PyDataStructs API. Like `LinearProbingMap` class in Python API and will have the same methods/attributes as `dict` but will use your C++ implementation. Can you give this idea a try?

PR #517, opened 2023-03-16
Implement method to check if string is present in Trie.

sak-codes:


pydatastructs/strings/trie.py
```diff
@@ -104,6 +104,24 @@ def is_present(self, string: str) -> bool:
...
+        =======
+        True if the given string is present in trie;
+        False in all other cases.
+        """
+        walk = self.root
+        for char in string:
+            if walk.get_child(char) is None:
+                return False
+            walk = walk.get_child(char)
+        return walk.is_terminal
```
czgdp1807, line 123: Seems like this means the string should be fully present in the trie? If its as a prefix then it will return `False`? Well, even if a string is present as a prefix it is still there in the trie.

pydatastructs/strings/tests/test_trie.py
```diff
@@ -2,14 +2,21 @@
...
         trie.insert(string)
 
+    prefix_strings = ["te", "t", "Am", "snb"]
+
     for string in strings:
         assert trie.is_present(string)
 
+    for string in prefix_strings:
+        assert trie.is_prefix_present(string)
+        assert not trie.is_present(string)
```
czgdp1807, line 18: Well, all the strings in `prefix_strings` are there in the trie. They are not absent, they are just present as prefixes of other strings.

czgdp1807: I think what you are trying to do can be achieved via adding a flag like, `is_inserted` in `TrieNode` and checking if its `True` in your method. I think similarly the method should be named, `is_inserted` and `is_present` should be kept untouched.

pydatastructs/strings/tests/test_trie.py
```diff
@@ -2,13 +2,20 @@
...
+    strings = ["A", "to", "tea", "ted", "ten", "i",
+               "in", "inn", "Amfn", "snbr"]
     trie = Trie()
     for string in strings:
         trie.insert(string)
 
+    prefix_strings = ["te", "t", "Am", "snb"]
+
     for string in strings:
+        assert trie.is_inserted(string)
```
czgdp1807, line 14: Let's also test by traversing `strings[::-1]` in a separate loop.

czgdp1807 (changes requested):

czgdp1807:
Seems like code quality tests are failing.

pydatastructs/strings/tests/test_trie.py
sak-codes: Done. Thank you.

czgdp1807 (approved):

pydatastructs/strings/tests/test_trie.py
```diff
@@ -23,7 +33,7 @@ def test_Trie():
         trie.delete(string)
         for present in strings:
             if present == string:
-                assert not trie.is_present(present)
+                assert not trie.is_inserted(present)
             else:
-                assert trie.is_present(present)
+                assert trie.is_inserted(present)
```
czgdp1807, line 38: Why call to is_present is removed? I think we should have both.

czgdp1807 (changes requested):

pydatastructs/strings/tests/test_trie.py
Smit-create: I think it's because the previous tests were actually meant to test `is_inserted`, and the bug is caught when we use `is_present`.
For example, if we have already deleted `"A"` from the trie, then still `trie.is_present("A")` will be `true` which is logically correct since `"Amfn"` is present. But the previous tests had this bug so I think this looks fine.

PR #521, opened 2023-03-25
Implement groups in DSU

sak-codes:

### DIFF
--- a/pydatastructs/miscellaneous_data_structures/disjoint_set.py
+++ b/pydatastructs/miscellaneous_data_structures/disjoint_set.py
@@ -130,12 +130,8 @@ class DisjointSetForest(object):
         Get the disjoint sets in DSU
         """
         result = dict()
-        for key, node in self.tree.items():
-            if node.parent == node:
-                parent = key
-            else:
-                parent = self.find_root(key)
-
+        for key in self.tree.keys():
+            parent = self.find_root(key).key
             members = result.get(parent, [])
             members.append(key)
             result[parent] = members
--- a/pydatastructs/miscellaneous_data_structures/disjoint_set.py
+++ b/pydatastructs/miscellaneous_data_structures/disjoint_set.py
@@ -125,16 +125,16 @@ class DisjointSetForest(object):
 
         return self.find_root(key).size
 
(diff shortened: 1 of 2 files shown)


pydatastructs/miscellaneous_data_structures/disjoint_set.py
```diff
@@ -124,3 +124,24 @@ def find_size(self, key):
             raise KeyError("Invalid key, %s"%(key))
 
         return self.find_root(key).size
+
+    def groups(self):
```
czgdp1807, line 128: By `groups` you mean returning a list of disjoint sets formed so far in the data structure?

czgdp1807: If so I'd say use `sets` because they are actually sets.

pydatastructs/miscellaneous_data_structures/disjoint_set.py
```diff
@@ -124,3 +124,24 @@ def find_size(self, key):
...
+        """
+        Get the groups in DSU
+        """
+        result = dict()
+        for key, node in self.tree.items():
+            if node.parent == node:
+                parent = key
+            else:
+                parent = node.parent.key
```
czgdp1807, line 137: To be on a safe side, I'd recommend calling `find_root` so that we associate a given node with the correct root.

czgdp1807 (changes requested):

pydatastructs/miscellaneous_data_structures/disjoint_set.py
```diff
@@ -124,3 +124,20 @@ def find_size(self, key):
             raise KeyError("Invalid key, %s"%(key))
 
         return self.find_root(key).size
+
+    def disjoint_sets(self):
+        """
+        Get the disjoint sets in DSU
```
czgdp1807, line 130:
```suggestion
        Returns a list of disjoint sets in the data structure.
```

czgdp1807 (changes requested):
Other than this rest looks good.

czgdp1807 (approved):

PR #523, opened 2023-04-01
Added Z-function implementation

CarolLuca:
I added an implementation of the Z-function of a string (also known as the Z-algorithm).

### DIFF
--- a/pydatastructs/strings/algorithms.py
+++ b/pydatastructs/strings/algorithms.py
@@ -201,8 +201,13 @@ def _boyer_moore(text, query):
                 shift += max(1, j + 1)
     return positions
 
-def _z_function(string):
-    z_fct = DynamicOneDimensionalArray(int, len(string))
+def _z_function(text, query):
+    if query == "":
+        string = text
+    else:
+        string = query + '$' + text
+
+    z_fct = OneDimensionalArray(int, len(string))
 
     curr_pos = 1
     seg_left = 0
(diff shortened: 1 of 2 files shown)


czgdp1807:
See, https://app.codecov.io/gh/codezonediitj/pydatastructs/pull/523/blob/pydatastructs/strings/algorithms.py. L206 and L231 are not tested. When `query = ""`, do we even need to do any kind of processing. In KMP we just do, `return DynamicOneDimensionalArray(int, 0)` if, `text` or `query` is empty. Why can't we do that in your implementation?

CarolLuca:
> See, [app.codecov.io/gh/codezonediitj/pydatastructs/pull/523/blob/pydatastructs/strings/algorithms.py](https://app.codecov.io/gh/codezonediitj/pydatastructs/pull/523/blob/pydatastructs/strings/algorithms.py). L206 and L231 are not tested. When `query = ""`, do we even need to do any kind of processing. In KMP we just do, `return DynamicOneDimensionalArray(int, 0)` if, `text` or `query` is empty. Why can't we do that in your implementation?

I see that the case is treated. The _z_vector() function doesn't do this because the actual vector it is often used also without a query, but for the _z_function(), I see I implemented the case when the lengths of the strings are zero. If I'm wrong please correct me, I might have missed something.

czgdp1807 (approved):

PR #539, opened 2023-06-29
Add lazy segment tree

sak-codes:

### DIFF
--- a/pydatastructs/miscellaneous_data_structures/segment_tree.py
+++ b/pydatastructs/miscellaneous_data_structures/segment_tree.py
@@ -1,4 +1,5 @@
 from .stack import Stack
+from pydatastructs import OneDimensionalArray
 from pydatastructs.utils.misc_util import (TreeNode,
     Backend, raise_if_backend_is_not_python)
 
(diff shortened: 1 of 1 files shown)


czgdp1807 (changes requested):
Run same tests of `OneDimensionalArraySegmentTree` with the new one. Put all tests into common function and pass objects of `OneDimensionalArraySegmentTree` and `OneDimensionalArraySegmentTreeLazy` to the common function.

sak-codes:
I am trying to design the method parameters such that I can generalize the implementation but since lazy updates for each of the different applications have different implementation, I am unable to find a common design.

czgdp1807:
> lazy updates for each of the different applications have different implementation, I am unable to find a common design.

Share some examples so that we can figure something out.

sak-codes:
Suppose we are solving an range sum update problem where we want to increment all the elements in the range $[i, j]$ with some number `x`. This is okay and well generalized for `j != i`. But when we want to change some element at index $i$, then if we use the same above approach then we need to pass by `y=x-cur_value[i]`, then during the update we will have `cur_value[i] += y`. So user will get confuse when to use `y` and when to use `x`.

pydatastructs/miscellaneous_data_structures/segment_tree.py
```diff
@@ -223,3 +224,138 @@ def query(self, start, end):
...
+            if start != end:
+                lazy_node.left.data = self._func((lazy_node.left.data, value))
+                lazy_node.right.data = self._func((lazy_node.right.data, value))
+            return
+        mid = (start + end) // 2
+        self._update_range(node.left, lazy_node.left, start, mid, l, r, value)
+        self._update_range(node.right, lazy_node.right, mid + 1, end, l, r, value)
+        node.data = self._func((node.left.data, node.right.data))
+
+    def update_range(self, start, end, value):
```
czgdp1807, line 322:
```suggestion
    def update(self, start, end, value):
```

pydatastructs/miscellaneous_data_structures/segment_tree.py
```diff
@@ -223,3 +224,138 @@ def query(self, start, end):
...
+        if not self.is_ready:
+            raise ValueError("{} tree is not built yet. ".format(self) +
+                             "Call .build method to prepare the segment tree.")
+        self._update_range(self._root, self._lazy_node, 0, len(self._array) - 1,
+                           start, end, value)
+
+    def update(self, index, value):
```
czgdp1807, line 329:
```suggestion
    def __setitem__(self, index, value):
```

czgdp1807: Same for the parent class.

czgdp1807 (changes requested):
I think right now the case is `update` for a single value and `update_range` for a range. I think we discussed to use `__setitem__` to update a single value. However, for a range we will use `update`. So please do it.

sak-codes:
So you want me to rename `update_range` to `update`?

PR #559, opened 2024-06-04
C++ backend for Self Balancing Binary Tree

Kishan-Ved:
Related to: Google Summer of Code 2024, Kishan Ved

### Objective
Add a C++ backend for Self Balancing Binary Tree class.


czgdp1807:
Seems like `codecov` isn't working. But we will see to it later.

PR #575, opened 2025-01-14
Fixed DynamicOneDimensionalArray's length function

### LINKED ISSUE #557: BUG: DynamicOneDimensionalArray's size is not updated properly

BUG: DynamicOneDimensionalArray's size is not updated properly

#### Example of the problem

```
def test_DynamicOneDimensionalArray3():
    DODA = DynamicOneDimensionalArray
    A = DODA(int, 1)
    A.delete(1)
    print(A._size)
    print(str(A))
    A.append(2)
    print(str(A))
    print(A.size)
    A.append(3)
    print(str(A))
    print(A.size)
```

Output:
```
1
[]
['2']
1
['2', '3']
3
```
The final size should be 2, but it is 3.

Naveen-Pal:
Fixes #557 "

Added a length function to obtain correct length of Dynamic array

image

### DIFF
--- a/pydatastructs/linear_data_structures/arrays.py
+++ b/pydatastructs/linear_data_structures/arrays.py
@@ -421,6 +421,9 @@ class DynamicOneDimensionalArray(DynamicArray, OneDimensionalArray):
     @property
     def size(self):
         return self._size
+    
+    def __len__(self):
+        return self._num
 
     def __str__(self):
         to_be_printed = ['' for _ in range(self._last_pos_filled + 1)]


Kishan-Ved:
@czgdp1807 windows-latest, 3.8 and macos-latest, 3.10 CI has also encountered failures. Expand and check the run tests section. However the CI check passes. It should fail. Isn't this weird?

Kishan-Ved:
@Naveen-Pal Try looking at places where len() was used for DODA. We will have to fix it there too. Because if len() was used in some other algo/test, that is bound to fail as we have changed the answer of len()

PR #577, opened 2025-01-22
added sorting algorithms like quick sort and merge sort

Mayank-tech69:
Implemented Merge Sort and Quick Sort algorithms, which are widely used for sorting large datasets efficiently.

### DIFF
--- a/pydatastructs/sorting/MERGE SORT/algorithm.md
+++ b/pydatastructs/sorting/MERGE SORT/algorithm.md
@@ -1,0 +1,88 @@
+# Merge Sort Algorithm 
+
+## **Merge Sort**
+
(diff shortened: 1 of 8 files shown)


czgdp1807 (commented):
Thank you for the contributions but we have these algorithms already,

https://github.com/codezonediitj/pydatastructs/blob/ed8831543cb14b0a84cef3c5ffdac46522f61015/pydatastructs/linear_data_structures/algorithms.py#L66

https://github.com/codezonediitj/pydatastructs/blob/ed8831543cb14b0a84cef3c5ffdac46522f61015/pydatastructs/linear_data_structures/algorithms.py#L689

PR #581, opened 2025-02-09
Bug fix

Susmita331:
- Ensures `_last_pos_filled` correctly reflects the last occupied index
- Uses a single-pass `left` pointer approach for efficient in-place swaps


Kishan-Ved:
The CI fails because of assertion errors, these errors check if the output of the function is equal to a particular value (let's call this v). Now, because you have changed the function, the output of the function is not equal to v.

Please update v accordingly.

More information on the error is available here: https://github.com/codezonediitj/pydatastructs/actions/runs/13228756520/job/36922991475?pr=581

(Scroll down to the end of the run tests section to find the error)

Issue #475, opened 2022-01-07
multithreaded fibonacci series


dejokz:
#### Description of the problem
implementation of multi-threaded Fibonacci series with a method fib_multi with takes the parameter n. 

#### Example of the problem
>fib_multi(n=20)
4181
>fib_multi(n=7)
8

#### References/Other comments
CLRS: chapter 27: Multithreaded algorithms
Not sure under which file hierarchy should this python file come under? Perhaps a new directory can be created to facilitate multithreaded algorithms.

czgdp1807:
Can you explain how multithreading is helpful here? The fastest algorithm that I know of is O(log n) to generate (n + 1)-th and n-th Fibonacci numbers. 

The more important question is what is the applicability of Fibonacci numbers? In which domain their efficient implementation is required to solve a non-trivial problem? Any references?

Abekaesh:
Is this issue relating to implementation of fastest algorithm to find nth fibonacci series number?

mridul45:
I would be happy to contribute to the project as I find my skills a good fit for the issue . As a GSSOC 23 member I humbly ask to allow me to contribute to the issue.

Meghaa702:
I would be happy to contribute to the project as I find my skills a good fit for the issue .  Please allow me to contribute to the issue.

Kushagra9399:
Well, I think it is better to use threading here for faster and efficient calculation and threading will also take less space as compared to recursion approach.
Hello, I am Kushagra Prajapat, want to work on this project in GSOC 2025. It seems your project is interesting as it is more logically. Hence, I would like to work on this project.
Please allow me to contribute on this project and I will provide you the required code within two to three days.

Kushagra9399:
Please refer the above content. I have created fibonacci.py file which includes the code for multi-threaded fibonacci series and I have verified it and checked it twice before submitting. I have also created test for that as per the problem given. Please see that and tell me if anything to change or edit, I will do that. I really liked this project. I want to work in this project in GSOC 2025.
Thank You

Kushagra9399:
One thing also according to your example if we assume fib_multi(1) = fib_multi(2) = 1 then your examples are working but in the code I have assumed fib_multi(0)=fib_multi(1) = 1 and then the series follows.

Kushagra9399:
Pull request #593

PR #593, opened 2025-03-04
Solved issue #475 Added fibonacci.py and test_fibonacci.py

Kushagra9399:

Please refer the above content. I have created fibonacci.py file which includes the code for multi-threaded fibonacci series and I have verified it and checked it twice before submitting. I have also created test for that as per the problem given. Please see that and tell me if anything to change or edit, I will do that. I really liked this project. I want to work in this project in GSOC 2025.
One thing also according to your example if we assume fib_multi(1) = fib_multi(2) = 1 then your examples are working but in the code I have assumed fib_multi(0)=fib_multi(1) = 1 and then the series follows.

Well, I think it is better to use threading here for faster and efficient calculation and threading will also take less space as compared to recursion approach.
Hello, I am Kushagra Prajapat, want to work on this project in GSOC 2025. It seems your project is interesting as it is more logically. Hence, I would like to work on this project.
Please allow me to contribute on this project.
Thank You

### DIFF
--- a/oryx-build-commands.txt
+++ b/oryx-build-commands.txt
@@ -1,0 +1,2 @@
+PlatformWithVersion=Python 
+BuildCommands=conda env create --file environment.yml --prefix ./venv --quiet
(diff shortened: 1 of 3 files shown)


Kushagra9399:
Please review this PR
@Kishan-Ved

Kushagra9399:
@czgdp1807 Please review this PR. If there are some changes required then tell me I will do that.

Issue #226, opened 2020-03-28
Allow dynamic graphs with Adjacency Matrix


czgdp1807:
#### Description of the problem

With the merging of #225 Adjacency matrix will allow dynamic graphs. So, the following methods should be defined,

1. `add_edge`
2. `remove_edge`

Along with defining them, `prim` implementation of minimum spanning tree should work for these graphs.

Aakansha99:
Kindly assign this issue to me

czgdp1807:
We do not assign issues. You can work on it without being assigned.

PR #595, opened 2025-03-05
Allow adding and removing vertices

larissapoghosyan:
After #225 graph refactoring this change simply enables adding and removing vertices to dynamic graphs.
Addresses the feature request #226.

- [x] Added add_vertex implementation. This was trivial
- [x] Added remove_vertex implementation - This required careful cleanup of existing edges, matrix rows and columns, in reverse order.
- [x] Added simple tests covering  addition and removal of vertices, including scenarios when we need to remove a vertix which has edges.


larissapoghosyan:
Hey @czgdp1807 any tips or suggestions?

Issue #588, opened 2025-03-03
Windows CI fails on README.md changes


GauravSaini01:

Kishan-Ved:
Related to: https://github.com/codezonediitj/pydatastructs/pull/585

Kishan-Ved:
This issue has been solved. We can close this.

PR #604, opened 2025-03-08
FIX: Windows CI

Kishan-Ved:
Recently, the windows CI has been failing due to some reason that is beyond my conscious. Hence, let's remove the command that caused the fail:

`run: conda --version`

Removal of this shouldn't matter as such.


Kishan-Ved:
Fixes: https://github.com/codezonediitj/pydatastructs/issues/588

Kushagra9399:
Please check and merge #593

Issue #603, opened 2025-03-08
Radix Sort


asmit27rai:
### Add Radix Sort Algorithm

**Description:**
Radix Sort is a non-comparative sorting algorithm that processes integer numbers by grouping them by individual digits. It sorts numbers digit by digit, starting from the least significant digit (LSD) to the most significant digit (MSD) or vice versa. It uses counting sort as a subroutine to sort the digits.

**References:**
- [Radix Sort - Wikipedia](https://en.wikipedia.org/wiki/Radix_sort)

asmit27rai:
@Kishan-Ved I am working on it.
Pls assign this to me...

Kishan-Ved:
Hi, we don't assign issues to specific people here. If it's important, please feel free to make a PR.

Kushagra9399:
I have added the radix sort algorithm along with its test cases. And all the tests worked successfully. Please check it.

PR #619, opened 2025-03-10
Adding Radix Sort #603

Kushagra9399:
I have added algorithm for radix sort in algorithms.py and its tests in test_algorithms.py


Kushagra9399 (commented):
@czgdp1807 Please check it once and if there need some changes then tell me about that, I will correct that.

PR #624, opened 2025-03-10
feat: add longest alternating subsequence

### LINKED ISSUE #317: Longest Alternating Subsequence
### Description of the problem
Longest Alternating Subsequence is a problem of finding a subsequence of a given sequence in which the elements are in alternating order, and in which the sequence is as long as possible.

https://en.wikipedia.org/wiki/Longest_alternating_subsequence

FamALouiz:
#### References to issue
Resolves #317 

#### Brief description of what was implemented
- Added longest alternating subsequence 
- Added longest alternating subsequence tests 

#### Comments
The algorithm works in O(n) using a dp + greedy solution. It does not return the actualy subsequence; instead, it returns only the length.

### DIFF
--- a/pydatastructs/linear_data_structures/algorithms.py
+++ b/pydatastructs/linear_data_structures/algorithms.py
@@ -1852,7 +1852,7 @@ def intro_sort(array, **kwargs) -> Array:
 
         return array
 
-def longest_alternating_subsequence(array: OneDimensionalArray, **kwargs) -> OneDimensionalArray:
+def longest_alternating_subsequence(array: OneDimensionalArray, **kwargs) -> int:
     """
     Finds the longest alternating subsequence in the given array.
     An alternating sequence is one in which the elements are in alternating order,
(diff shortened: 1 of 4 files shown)


pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -1850,3 +1851,69 @@ def partition(array, lower, upper):
...
+        # Increasing changes if decreasing changes
+        if (array[i] > array[i-1]):
+            increasing = decreasing + 1
+
+        # Decreasing changes if increasing changes
+        elif (array[i] < array[i-1]):
+            decreasing = increasing + 1
+
+    # Return the maximum length
+    return max(increasing, decreasing)
```
czgdp1807, line 1919: Would be better to return the increasing sequence as well and not just the length.

czgdp1807 (changes requested):

FamALouiz:
@czgdp1807 Could you please review this? Changes have been made.

pydatastructs/linear_data_structures/algorithms.py
FamALouiz: Done.

PR #625, opened 2025-03-10
feat: Implement Yen's Algorithm for K-Shortest Paths

asmit27rai:
### 📌 Description  
This PR introduces **Yen's Algorithm**, which finds the **K shortest paths** between a source and destination node in a graph. The algorithm efficiently computes multiple alternative paths, making it useful for applications such as:  
- **Network Routing** (finding alternative paths in case of congestion)  
- **Transportation Planning** (evaluating different routes)  
- **Logistics Optimization** (determining the best delivery paths)  

### 🛠️ Changes  
- Implemented Yen's Algorithm in the graph module.  
- Added helper functions to extract deviations and manage path modifications.  
- Included test cases to validate correctness and efficiency.

### DIFF
--- a/pydatastructs/graphs/algorithms.py
+++ b/pydatastructs/graphs/algorithms.py
@@ -11,6 +11,8 @@ from pydatastructs.miscellaneous_data_structures import (
 from pydatastructs.graphs.graph import Graph
 from pydatastructs.linear_data_structures.algorithms import merge_sort_parallel
 from pydatastructs import PriorityQueue
+from copy import deepcopy
+import heapq
 
 __all__ = [
     'breadth_first_search',
@@ -23,7 +25,8 @@ __all__ = [
     'all_pair_shortest_paths',
     'topological_sort',
     'topological_sort_parallel',
-    'max_flow'
+    'max_flow',
+    'yen_algorithm'
 ]
 
 Stack = Queue = deque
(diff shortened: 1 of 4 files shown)


pydatastructs/graphs/algorithms.py
```diff
@@ -11,6 +11,8 @@
 from pydatastructs.graphs.graph import Graph
 from pydatastructs.linear_data_structures.algorithms import merge_sort_parallel
 from pydatastructs import PriorityQueue
+from copy import deepcopy
+import heapq
```
czgdp1807, line 15: We do have `heapq` implemented in PyDataStructs? I think `PriorityQueue` and `heapq` are the same?

pydatastructs/graphs/tests/test_algorithms.py
```diff
@@ -2,8 +2,9 @@
 breadth_first_search_parallel, minimum_spanning_tree,
 minimum_spanning_tree_parallel, strongly_connected_components,
 depth_first_search, shortest_paths,all_pair_shortest_paths, topological_sort,
-topological_sort_parallel, max_flow)
+topological_sort_parallel, max_flow, yen_algorithm)
 from pydatastructs.utils.raises_util import raises
+import pytest
```
czgdp1807, line 7: We don't need to import this.

pydatastructs/graphs/tests/test_algorithms.py
```diff
@@ -448,3 +449,40 @@ def _test_max_flow(ds, algorithm):
...
+            ['V1', 'V2', 'V3', 'V5'],
+            ['V1', 'V3', 'V5'],
+            ['V1', 'V2', 'V3', 'V4', 'V5']
+        ]
+
+        assert len(k_shortest_paths) == 3, "Expected 3 shortest paths"
+        for i, path in enumerate(k_shortest_paths):
+            assert path == expected_paths[i], f"Path {i} does not match expected path. Got {path}, expected {expected_paths[i]}"
+
+    _test_yen_algorithm("List")
```
czgdp1807, line 488: Why don't we support matrix representation of graphs?

asmit27rai: > Why don't we support matrix representation of graphs?

This means that the AdjacencyMatrix implementation does not support dynamic modifications like removing vertices or edges during runtime. Since Yen's algorithm requires modifying the graph (e.g., removing edges and vertices to find alternative paths), it cannot work with the AdjacencyMatrix implementation in its current form.

Issue #628, opened 2025-03-11
Add queue improved Bellman-Ford algorithm


TripleCamellya:
Add the queue version of the Bellman-Ford algorithm.

Also known as the “shortest path faster algorithm”.

For the Bellman-Ford algorithm, each iteration requires relaxing all edges. For a vertex v that has a distance value that has not changed since the last time the edges out of v were relaxed, then there is no need to relax the edges out of v a second time. 

Keep a collection of vertices whose outgoing edges need to be relaxed, reducing the number of vertexs to be relaxed.

https://en.wikipedia.org/wiki/Bellman%E2%80%93Ford_algorithm#Improvements

TripleCamellya:
Update the original Bellman Ford algorithm with queue version.

PR #629, opened 2025-03-11
feat: replace Bellman Ford Algorithm with queue version

TripleCamellya:
Implementing queue improved bellman ford algorithm.

Also known as the “shortest path faster algorithm”.

For the Bellman-Ford algorithm, each iteration requires relaxing all edges. For a vertex v that has a distance value that has not changed since the last time the edges out of v were relaxed, then there is no need to relax the edges out of v a second time.

Changes in PR:
1. Implementation of queue improved bellman ford algorithm.
2. Add tests for queue improved bellman ford algorithm.
3. Add choices for shorted path algorithm and related comment.

Related to https://github.com/codezonediitj/pydatastructs/issues/628

### DIFF
--- a/pydatastructs/graphs/algorithms.py
+++ b/pydatastructs/graphs/algorithms.py
@@ -697,7 +697,7 @@ def shortest_paths(graph: Graph, algorithm: str,
         The algorithm to be used. Currently, the following algorithms
         are implemented,
 
-        'bellman_ford' -> Bellman-Ford algorithm as given in [1], with a queue to improve performance on sparse graphs.
+        'bellman_ford' -> Bellman-Ford algorithm as given in [1]
 
         'dijkstra' -> Dijkstra algorithm as given in [2].
     source: str
--- a/pydatastructs/graphs/algorithms.py
+++ b/pydatastructs/graphs/algorithms.py
@@ -697,11 +697,9 @@ def shortest_paths(graph: Graph, algorithm: str,
         The algorithm to be used. Currently, the following algorithms
         are implemented,
 
-        'bellman_ford' -> Bellman-Ford algorithm as given in [1].
+        'bellman_ford' -> Bellman-Ford algorithm as given in [1], with a queue to improve performance on sparse graphs.
 
         'dijkstra' -> Dijkstra algorithm as given in [2].
-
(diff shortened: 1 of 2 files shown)


pydatastructs/graphs/algorithms.py
```diff
@@ -700,6 +700,8 @@ def shortest_paths(graph: Graph, algorithm: str,
         'bellman_ford' -> Bellman-Ford algorithm as given in [1].
 
         'dijkstra' -> Dijkstra algorithm as given in [2].
+
+        'queue_improved_bellman_ford' -> Queue Improved Bellman-Ford algorithm as given in [3].
```
czgdp1807, line 704: Instead of a new one, let's just use Queues in Bellman Ford algorithm itself? Any specific for not doing so?

czgdp1807: If Queue version is always better then use Queue only. What's the difference between running times with and without queues?

czgdp1807: So let's always use a queue. Change the original bellman ford implementation.

pydatastructs/graphs/algorithms.py
```diff
@@ -697,7 +697,7 @@ def shortest_paths(graph: Graph, algorithm: str,
         The algorithm to be used. Currently, the following algorithms
         are implemented,
 
-        'bellman_ford' -> Bellman-Ford algorithm as given in [1].
+        'bellman_ford' -> Bellman-Ford algorithm as given in [1], with a queue to improve performance on sparse graphs.
```
czgdp1807, line 700:
```suggestion
        'bellman_ford' -> Bellman-Ford algorithm as given in [1]
```

Implementation detail aren't required to be mentioned.

czgdp1807 (changes requested):

czgdp1807:
@TripleCamellya Since, this PR is going to go in. Let's take out time and meet. What's your email ID?

TripleCamellya:
> @TripleCamellya Since, this PR is going to go in. Let's take out time and meet. What's your email ID?

Thanks! My email is acrossthewallfhh@gmail.com.

Issue #634, opened 2025-03-12
Implement `add_vertex` and `remove_vertex` for Dynamic Graphs in `AdjacencyMatrix`


asmit27rai:
#### Description
Currently, the `AdjacencyMatrix` class in `pydatastructs` only supports static graphs, as indicated by the `NotImplementedError` in the `add_vertex` and `remove_vertex` methods. This limitation restricts the use of the adjacency matrix representation for dynamic graphs where vertices can be added or removed after the graph is initialized.

#### Expected Behavior
The `add_vertex` and `remove_vertex` methods should be implemented to allow dynamic modification of the graph. Specifically:
1. **`add_vertex`**: This method should add a new vertex to the graph and update the adjacency matrix and edge weights accordingly.
2. **`remove_vertex`**: This method should remove a vertex from the graph, along with all edges connected to it, and update the adjacency matrix and edge weights.

#### Proposed Solution
1. **`add_vertex`**:
   - Add the new vertex to the `vertices` list.
   - Update the `matrix` dictionary to include the new vertex as a key with an empty dictionary as its value.
   - Ensure that the new vertex is added as a column in the adjacency matrix for all existing vertices.

2. **`remove_vertex`**:
   - Remove the vertex from the `vertices` list.
   - Remove the vertex from the `matrix` dictionary and all its associated edges.
   - Remove any edge weights involving the vertex from the `edge_weights` dictionary.

asmit27rai:
@Kishan-Ved @czgdp1807 I am working on this issue..

PR #640, opened 2025-03-12
feat: Implement add_vertex and remove_vertex for AdjacencyMatrix

asmit27rai:
## Description

This PR implements the `add_vertex` and `remove_vertex` methods for the `AdjacencyMatrix` class, enabling support for dynamic graphs. Previously, the adjacency matrix implementation only supported static graphs, as indicated by the `NotImplementedError` in these methods. This enhancement allows users to add or remove vertices after the graph is initialized, making the adjacency matrix representation more versatile.

## Changes

- **`add_vertex` Method**:
  - Adds a new vertex to the `vertices` list.
  - Initializes a new row and column in the adjacency matrix for the new vertex.
  - Sets all entries in the new row and column to `False` (indicating no edges initially).

- **`remove_vertex` Method**:
  - Removes the vertex from the `vertices` list.
  - Removes the corresponding row and column from the adjacency matrix.
  - Removes all edges connected to the vertex from the `edge_weights` dictionary.

- **Tests**:
  - Added test cases to verify the functionality of `add_vertex` and `remove_vertex`.
  - Ensured that the adjacency matrix and edge weights are updated correctly.

## Fixes : #634

### DIFF
--- a/pydatastructs/graphs/adjacency_matrix.py
+++ b/pydatastructs/graphs/adjacency_matrix.py
@@ -57,12 +57,30 @@ class AdjacencyMatrix(Graph):
         return neighbors
 
     def add_vertex(self, node):
-        raise NotImplementedError("Currently we allow "
-                "adjacency matrix for static graphs only")
+        node_name = str(node.name)
+        if node_name in self.vertices:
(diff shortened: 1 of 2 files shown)


asmit27rai:
@czgdp1807 Pls review this PR

PR #642, opened 2025-03-13
feat: Add Radix Sort Algorithm

asmit27rai:
### 📌 Description
This PR introduces the **Radix Sort** algorithm, a non-comparative sorting technique that processes numbers digit by digit. It utilizes **Counting Sort** as a subroutine to achieve stable sorting. The sorting process starts from the **Least Significant Digit (LSD)** and moves towards the **Most Significant Digit (MSD)**.

### 📖 References
- [Radix Sort - Wikipedia](https://en.wikipedia.org/wiki/Radix_sort)

# FIXES : #603

### DIFF
--- a/pydatastructs/linear_data_structures/algorithms.py
+++ b/pydatastructs/linear_data_structures/algorithms.py
@@ -30,7 +30,8 @@ __all__ = [
     'jump_search',
     'selection_sort',
     'insertion_sort',
-    'intro_sort'
+    'intro_sort',
+    'radix_sort',
 ]
 
 def _merge(array, sl, el, sr, er, end, comp):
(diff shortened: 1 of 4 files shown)


pydatastructs/linear_data_structures/algorithms.py
```diff
@@ -1850,3 +1851,71 @@ def partition(array, lower, upper):
...
+    Parameters
+    ==========
+    array: Array
+        The array which is to be sorted.
+    comp: lambda/function
+        The comparator which is to be used
+        for sorting. Optional, by default, less than or
+        equal to is used for comparing two
+        values.
+    """
```
czgdp1807, line 1867:
```suggestion
    """
    Implements Radix Sort.

    Parameters
    ==========

    array: Array
        The array which is to be sorted.
    comp: lambda/function
        The comparator which is to be used
        for sorting. Optional, by default, less than or
        equal to is used for comparing two
        values.
    """
```

czgdp1807 (changes requested):

pydatastructs/linear_data_structures/algorithms.py
asmit27rai: @czgdp1807 Done as per requested....

PR #646, opened 2025-03-15
Added Bipartite Algorithm

### LINKED ISSUE #645: Add Bipartite Graph Checking Algorithm
**Description:**

We can add a new algorithm to PyDataStructs that checks whether a given undirected graph is bipartite. A graph is bipartite if its vertices can be divided into two disjoint sets such that no two vertices within the same set are adjacent. This property is crucial in many applications, including scheduling, matching, and network flow. The algorithm will use a BFS-based approach to assign two colors to vertices and detect any conflicts, running in O(V + E) time.

**Tasks:**

1. Implement the function for checking graph bipartiteness.
2. Write test cases to verify its correctness for both adjacency list and adjacency matrix implementations.
3. Update the documentation.

**References:**

https://en.wikipedia.org/wiki/Bipartite_graph

hargunkaur286:

Fixes #645 

This PR introduces the implementation of a bipartite-check algorithm.

1. Added the is_bipartite function in pydatastructs/graphs/algorithms.py to determine whether an undirected graph is bipartite using a BFS-based approach. The algorithm runs in O(V + E) time and returns a tuple containing a boolean indicating bipartiteness and a dictionary mapping vertices to their assigned colors.

2. Updated the documentation with an example usage snippet, and relevant reference link.

3. Updated the __all__ list to expose is_bipartite as part of the public API.

4. Added the test cases for the bipartite-check algorithm.

### DIFF
--- a/pydatastructs/graphs/__init__.py
+++ b/pydatastructs/graphs/__init__.py
@@ -21,7 +21,8 @@ from .algorithms import (
     all_pair_shortest_paths,
     topological_sort,
     topological_sort_parallel,
-    max_flow
+    max_flow,
+    is_bipartite
 )
 
 __all__.extend(algorithms.__all__)
--- a/pydatastructs/graphs/algorithms.py
+++ b/pydatastructs/graphs/algorithms.py
@@ -23,7 +23,8 @@ __all__ = [
     'all_pair_shortest_paths',
     'topological_sort',
     'topological_sort_parallel',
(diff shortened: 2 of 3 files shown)


czgdp1807:
Thanks for this. Actually there is another PR for this - https://github.com/codezonediitj/pydatastructs/pull/655/files and it looks better to me. So closing this one.

Issue #650, opened 2025-03-17
Raita Algorithm for String Matching


asmit27rai:
### Implement Raita Algorithm for String Matching

**Description:**
The Raita algorithm is a variation of the Boyer-Moore algorithm that uses a different heuristic for shifting the pattern. It is efficient for single pattern matching and has an average-case complexity of O(n + m), where `n` is the length of the text and `m` is the length of the pattern.

This issue involves implementing the Raita algorithm in the `pydatastructs.strings.algorithms` module and adding corresponding test cases.

asmit27rai:
@Kishan-Ved I am working on this issue..

PR #651, opened 2025-03-17
feat: Implement Raita Algorithm for string matching

asmit27rai:
# Implement Raita Algorithm for String Matching

## Description
This PR adds the **Raita Algorithm** to the `pydatastructs.strings.algorithms` module. The Raita algorithm is a variation of the Boyer-Moore algorithm and is efficient for single pattern matching with an average-case complexity of **O(n + m)**.

## Changes
1. Added the `_raita` function to `pydatastructs/strings/algorithms.py`.
2. Updated the `find` function to include the `raita` algorithm.
3. Added test cases for the Raita algorithm in `tests/test_strings.py`.

Fixes: #650

### DIFF
--- a/pydatastructs/strings/algorithms.py
+++ b/pydatastructs/strings/algorithms.py
@@ -245,3 +245,48 @@ def _z_function(text, query):
             positions.append(pos)
 
     return positions
+
+def _raita(text, query):
+    """
+    Implements the Raita algorithm for string matching.
+
+    Parameters
+    ==========
+    text: str
+        The text in which the pattern is to be searched.
+    query: str
+        The pattern to be searched in the text.
+
+    Returns
+    =======
+    DynamicOneDimensionalArray
+        An array of starting positions of the pattern in the text.
+    """
+    positions = DynamicOneDimensionalArray(int, 0)
+    n, m = len(text), len(query)
+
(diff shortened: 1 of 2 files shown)


asmit27rai:
@Kishan-Ved @czgdp1807 Pls review this PR...

pydatastructs/strings/algorithms.py
```diff
@@ -245,3 +245,48 @@ def _z_function(text, query):
...
+    text: str
+        The text in which the pattern is to be searched.
+    query: str
+        The pattern to be searched in the text.
+
+    Returns
+    =======
+    DynamicOneDimensionalArray
+        An array of starting positions of the pattern in the text.
+    """
```
czgdp1807, line 264:
```suggestion
    Parameters
    ==========

    text: str
        The text in which the pattern is to be searched.
    query: str
        The pattern to be searched in the text.

    Returns
    =======

    DynamicOneDimensionalArray
        An array of starting positions of the pattern in the text.
    """
```

czgdp1807 (changes requested):

pydatastructs/strings/algorithms.py
asmit27rai: @czgdp1807 Done with the changes..

PR #662, opened 2025-03-19
feat: add Parent Pointer Tree

### LINKED ISSUE #325: Parent pointer tree
In computer science, an in-tree or parent pointer tree is an N-ary tree data structure in which each node has a pointer to its parent node, but no pointers to child nodes.

#### Example of the problem
https://opendsa-server.cs.vt.edu/ODSA/Books/Everything/html/UnionFind.html

https://en.wikipedia.org/wiki/Parent_pointer_tree

FamALouiz:
#### References to Issue
Resolves #325 

#### Brief description of what is implemented 
- Added parent pointer tree for python backend
- Added parent pointer tree tests

### DIFF
--- a/pydatastructs/trees/m_ary_trees.py
+++ b/pydatastructs/trees/m_ary_trees.py
@@ -171,7 +171,7 @@ class MAryTree(object):
                         to_be_printed[i].append(j)
         return str(to_be_printed)
 
-class ParentPointerTree(object):
+class ParentPointerTree(MAryTree):
     """
     Implements a tree with parent pointers.
 
@@ -296,7 +296,7 @@ class ParentPointerTree(object):
         key
             The key for searching.
         parent: bool
-            If true then returns index of the
+            If true then returns node of the
             parent of the node with the passed
             key.
             By default, False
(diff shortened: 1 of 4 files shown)


TripleCamellya:
In my opnion, for add a new data structure or algorithm for public, the final interception is reserved by czgdp1807.

First step is append the class or function in the beginning array `__all__` of the file. Like:
```python
__all__ = [
    'MAryTree',
    'ParentPointerTree'
]
```

Secondly, add the API in the module's `__init__.py` file. Like
```python
from .m_ary_trees import (
    MAryTreeNode, MAryTree,
    ParentPointerTreeNode, ParentPointerTree
)
```

For `ParentPointerTreeNode`, you can check the above steps.

Three, after above two steps, you can import `ParentPointerTree` by

```python
from pydatastructs import MAryTree, ParentPointerTree
```

Lastly, update your new datastructure in docs.

FamALouiz:
Yea, points 1, 2, and 4 are valid. I had forgotten to add them. 

As for point 3, I was following the importing style in the other files. In the other files, they important it differently. For example,
```python
from pydatastructs.linear_data_structures.arrays import ArrayForTrees
```

However, The points are nice regardless thanks :)

PR #680, opened 2025-04-08
added c++ backend for cocktail sort

### LINKED ISSUE #679: Implement C++ Backend for Cocktail Sort
This issue proposes adding a C++ backend for the Cocktail Sort algorithm (also known as Bidirectional Bubble Sort or Shaker Sort) to the pydatastructs library. Cocktail Sort is a comparison-based sorting algorithm that improves upon Bubble Sort by sorting an array or list in both directions—forward (left to right) and backward (right to left)—in alternating passes. This bidirectional approach reduces the number of passes needed for already partially sorted data, making it more efficient in certain cases.

arvinder004:
Fixes #679

- Added the cocktail sort implementation in the C++ Backend file `quadratic_time_sort.hpp`
- Exposed C++ functions to Python
- Updated Python Functions

### DIFF
--- a/pydatastructs/linear_data_structures/_backend/cpp/algorithms/algorithms.cpp
+++ b/pydatastructs/linear_data_structures/_backend/cpp/algorithms/algorithms.cpp
@@ -4,6 +4,8 @@
 #include "misc_algorithms.hpp"
 
 static PyMethodDef algorithms_PyMethodDef[] = {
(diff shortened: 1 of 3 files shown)


arvinder004:
sir i have added a test function, it was not accepting the `_test_common_sort()` so I have added a seperate test function for cocktail shaker sort that will test both python and C++ Backends

PR #684, opened 2025-06-14
Add C++ Backend for BFS and Use std::variant for Graph Node Data

prex03:
- Added C++ implementations for DFS and BFS to improve traversal performance.
- Updated graph node data handling to use std::variant for type-safe storage.
- Adjusted Python bindings to support these backend changes.


prex03:
Tests for PyObject* type of data for nodes added in test_misc_util.py lines 39 and 47

czgdp1807 (approved):

